PR #11, opened 2022-05-31
Initial documentation

### LINKED ISSUE #7: Add .tex for masterdoc.pdf
[masterdoc.pdf ](https://github.com/UCL/isotropic-finite-difference-time-domain/blob/main/masterdoc.pdf) was generated with latex rather than pdflatex. The tex file needs to be updated to be compilable, and added to the repository. It is also lacking information about some features in the code.

t-young31:
Closes #7 

Adds the .tex documentation and a GitHub actions workflow to automatically build and commit it to a `gh-doc` branch.

***
### Dependencies

- #2 
> Please don't merge until these have been merged

*** 
### TODO

- [x] Add README with instructions for compiling
- [x] Add references bib
- [x] Add doc CI

*** 
### Reviewers

Given the very large diff which comes from the addition of `masterdoc.tex` it's perhaps worth just focusing on the other files that have been modified. Specific questions:

- Is the README clear? 
- Does it compile on your machine?


giordano (commented):
Couple of comments:

* I think there are unrelated changes coming from #2
* I verified locally documentation can be built successfully by running `make -C doc`!  However I'd suggest removing the generated PDF now.

t-young31:
Thanks for the Makefile suggestions 😄 

> Also, since now the document can be automatically built from source, I'd actually remove it from the repository, otherwise everybody recompiling the document locally could accidentally push it by accident, creating lots of unnecessary binary diffs, which only bloat the repo. We could for example have automatic CI/CD if we want to serve the PDF somewhere else (a separate branch is an option, for example).

Definitely would be nice to to avoid potential binary diffs – building automatically and linking to a branch for the docs would be nice, but for the time being we could just remove `cp $(BUILD_DIR)/$@ $@`?

doc/Makefile
```diff
@@ -0,0 +1,18 @@
+.PHONY: pdf clean
+
+BUILD_DIR = build
+DOC_FILE = masterdoc.pdf
+
+pdf: $(DOC_FILE)
+
+$(DOC_FILE): masterdoc.tex refs.bib $(wildcard figures/*.pdf)
+	mkdir -p $(BUILD_DIR)
+	cp refs.bib $(BUILD_DIR)/.
```
giordano, line 10: Min nitpick: I just realised there is no need to copy this file
```suggestion
```

giordano (commented):
> Definitely would be nice to to avoid potential binary diffs – building automatically and linking to a branch for the docs would be nice,

For example some years ago I read https://dfm.io/posts/travis-latex/, which outlines a good overall strategy (I'm not sure I agree with using Tectonic, I'd go with TeX Live and pdflatex), although today we'd likely want to do this with GitHub Actions.

> but for the time being we could just remove `cp $(BUILD_DIR)/$@ $@?`

No, I think that's fine, it's nice to have the file in the current `doc` directory for who builds it locally, but I'd still remove the file from the repository and add it to `.gitignore`.

t-young31:
pdf is currently built at [init_doc/doc/masterdoc.pdf](https://github.com/t-young31/ifdtd/blob/gh-doc/doc/masterdoc.pdf) so the badge link is broken, but it _should_ work when merged into main. Thoughts appreciated!

.github/workflows/doc.yml
```diff
@@ -0,0 +1,26 @@
+name: Build LaTeX documentation
+on: [push]
```
giordano, line 2: We probably want to restrict to pushes to `main`?  Otherwise also pushes to `gh-doc` (and any other branch) would trigger this, causing an interesting loop.

.github/workflows/doc.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  build_latex:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up repository
+        uses: actions/checkout@v2
```
giordano, line 8: There is version 3 now:
```suggestion
        uses: actions/checkout@v3
```

.github/workflows/doc.yml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          ref: ${{ github.head_ref }}
```
giordano, line 10: Not sure these lines are needed?

giordano (commented):
I haven't studied the workflow very well, but it looks ok, I guess we'll see in it in action after it's merged, and fix it if necessary (hopefully not!).  Nice work overall!!

Only remaining issue is that this PR is [still pulling changes](https://github.com/UCL/TDMS/pull/11#pullrequestreview-1004795910) from #2 :slightly_smiling_face:

.github/workflows/doc.yml
t-young31: I did mean to do this – left as `on [push]` so I could test the [action](https://github.com/t-young31/ifdtd/actions/runs/2516126147), but forgot to change it back. thanks 👍🏼

.github/workflows/doc.yml
t-young31: They were in the example action for one of the stages, so left as is

giordano:
Looks like you removed the `.gitignore` by accident in the [forced push](https://github.com/UCL/TDMS/compare/a5d54a289cb64ab1a0517ea6997cad63db8c452b..2ea269a94405cb4b9cde23f402e99661373e879a).

t-young31:
That's because it's in #2 😅

giordano (approved):
:rocket:

PR #27, opened 2022-06-21
Update workflow

t-young31:
- Updates the workflow to clean the branch other than the .pdf
- Fixes the badge link

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -18,9 +18,10 @@ jobs:
         with:
           root_file: masterdoc.tex
           working_directory: doc
-      - name: Move pdf
+      - name: Move pdf and remove other files
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/doc.yml
```diff
@@ -18,9 +18,10 @@ jobs:
         with:
           root_file: masterdoc.tex
           working_directory: doc
-      - name: Move pdf
+      - name: Move pdf and remove other files
         run: |
           mv doc/masterdoc.pdf .
+          ls | grep -xv "masterdoc.pdf" | xargs rm -r
```
giordano, line 24: You may need forcing removing
```suggestion
          ls | grep -xv "masterdoc.pdf" | xargs rm -rf
```

giordano (approved):

PR #29, opened 2022-07-07
Add CI setup for GitHub Actions

giordano:
Continuous Integration (CI) allows us to automatically run tests for all the changes that are proposed in pull requests, and when they are pushed to the `main` branch, so that we make sure we don't accidentally break anything (as long as it is tested).


t-young31:
Looks awesome – thanks 🎉 . I'll have a proper look when I'm back from holiday on Thursday (14th)

t-young31 (approved):

PR #31, opened 2022-07-14
Move system tests and update CI workflow

t-young31:
- Move system tests to /tdms/tests/system
- Update GitHub actions CI workflow to execute two system tests
- Adds python requirements.txt

### Dependencies 
- #32


tdms/src/tdms_iterator.cpp
```diff
@@ -6087,28 +6090,31 @@ if(runmode == rm_complete && (nvertices>0) )
   }
   
   //now find the maximum absolute value of residual field in the grid
-  for(k=0;k<(K_tot+1);k++)
```
giordano, line 6090: Reminder that this includes changes from #32, so we should probably wait for that PR to be merged before going on with this one.

tdms/tests/system/utils.py
```diff
@@ -38,7 +29,7 @@ class HDF5File(dict):
 
     COMPLEX_TYPE_STRING = r"[('real', '<f8'), ('imag', '<f8')]"
 
-    def __init__(self, filepath: Path):
+    def __init__(self, filepath: Union[str, Path]):
```
giordano, line 32: Out of curiosity, are these type annotations actually checked at any point?

tdms/tests/system/utils.py
```diff
@@ -75,6 +66,11 @@ def matches(self, other: "HDF5File", rtol=1e-10) -> bool:
 
         for key, value in self.items():
 
+            if key == "maxresfield":
+                other_value = other[key]
+                print(value, other_value, relative_mean_squared_difference(value, other_value))
+                continue  # TODO: remove
```
giordano, line 72: Why we can't remove the `continue` now?

tdms/CMakeLists.txt
```diff
@@ -71,7 +71,9 @@ target_sources(tdms PRIVATE
 
 target_link_libraries(tdms
         FFTW::Double
-        ${Matlab_LIBRARIES}
+        ${Matlab_MEX_LIBRARY}
+        ${Matlab_MX_LIBRARY}
+        ${Matlab_MAT_LIBRARY}
```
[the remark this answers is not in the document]
t-young31: Failed to run on Myriad with 2019 without it 😞

tdms/src/tdms_iterator.cpp
```diff
@@ -336,7 +336,7 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
   //these are used for boot strapping. There is currently no way of exporting this.
   double **iwave_lEx_Rbs, **iwave_lEy_Rbs, **iwave_lHx_Rbs, **iwave_lHy_Rbs,**iwave_lEx_Ibs, **iwave_lEy_Ibs, **iwave_lHx_Ibs, **iwave_lHy_Ibs;
   double *to_l, *hwhm, *omega_an, *dt;
-  double maxfield = 0, tempfield;
+  double maxfield = 0.0, tempfield;
```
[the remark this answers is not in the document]
t-young31: agree 👍🏼  (changed while testing)

tdms/src/tdms_iterator.cpp
t-young31: 💯

tdms/tests/system/utils.py
t-young31: Not checked, just highlighted by PyCharm

tdms/tests/system/utils.py
t-young31: we can – I just forgot to 🤦🏼

tdms/CMakeLists.txt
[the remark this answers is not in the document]
t-young31: indeed 🤦🏼

giordano:
Can you please rebase on `main`?

tdms/src/tdms_iterator.cpp
```diff
@@ -5940,28 +5940,31 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
...
-	      maxfield = tempfield;
-	    tempfield = fabs(Hzx[k][j][i]+Hzy[k][j][i]);	
-	    if( maxfield < tempfield )
-	      maxfield = tempfield;
-	  }
+      for(k=0;k<(K_tot+1);k++) {
+        for (j = 0; j < (J_tot + 1); j++){
+          for (i = 0; i < (I_tot + 1); i++) {
+            tempfield = fabs(Exy[k][j][i] + Exz[k][j][i]);
+            if (maxfield < tempfield)
```
giordano, line 5947: Since you're modifying also this block, could add the brackets also to these `if`s

tdms/tests/system/utils.py
```diff
@@ -100,3 +91,53 @@ def relative_mean_squared_difference(a: np.ndarray, b: np.ndarray) -> float:
...
+                shutil.rmtree(dir_path)
+
+            return result
+
+        return wrapped_function
+
+    return func_decorator
+
+
+def run_tdms(input_filename: str, output_filename: str) -> None:
```
giordano, line 132: Curiosity No. 2 (not really important): is `None` the type of `None`?  Or I misunderstand that the `->` syntax indicates the _type_ of the output of the function?

giordano (approved):

tdms/tests/system/utils.py
t-young31: Interesting.. the `->` should be the type of the output but it's I believe convention to have `None` for a void function

```python
>>> type(None)
<class 'NoneType'>
```

but it can't be instantiated – I guess having only one None is a good thing

PR #32, opened 2022-07-14
Reset the maxfield to zero

t-young31:
Solves a bug initially thought to be Mac-specific but was the cause of flaky system tests (see #31)

The `maxfield` updated in the loop could exceed all the `tempfield` values, meaning that it was not updated upon loop completion. This PR simply resets `maxfield` to zero (and adds some explicit for loop scopes).

### DIFF
--- a/tdms/src/tdms_iterator.cpp
+++ b/tdms/src/tdms_iterator.cpp
@@ -6087,28 +6087,31 @@ if(runmode == rm_complete && (nvertices>0) )
   }
   
   //now find the maximum absolute value of residual field in the grid
-  for(k=0;k<(K_tot+1);k++)
-    for(j=0;j<(J_tot+1);j++)
-      for(i=0;i<(I_tot+1);i++){
-	tempfield = fabs(Exy[k][j][i]+Exz[k][j][i]);
-	if( maxfield < tempfield )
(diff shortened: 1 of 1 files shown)


giordano (approved):
Looks good as far as I can tell, but I'm not intimately familiar with this code.

tdms/src/tdms_iterator.cpp
```diff
@@ -6087,28 +6087,38 @@ if(runmode == rm_complete && (nvertices>0) )
...
-	if( maxfield < tempfield )
-	  maxfield = tempfield;
-	tempfield = fabs(Hyx[k][j][i]+Hyz[k][j][i]);	
-	if( maxfield < tempfield )
-	  maxfield = tempfield;
-	tempfield = fabs(Hzx[k][j][i]+Hzy[k][j][i]);	
-	if( maxfield < tempfield )
-	  maxfield = tempfield;
+  // after resetting the maxfield value calculated in the main loop
+  maxfield = 0.0;
```
prmunro, line 6091: Good pickup, this needs to be here to be correct under all circumstances. In some cases the code will behave the same since maxfield is reset to 0 in the main loop, but only on certain iterations.

PR #34, opened 2022-07-15
Add additional system tests

t-young31:
Adds additional system tests. Due to their large size the input and reference output have been uploaded to [Zenodo](https://zenodo.org/) ([arc_01](https://zenodo.org/record/6838866#.YtFemezMKrM), [arc_02](https://zenodo.org/record/6838977#.YtFepezMKrM), [arc_03](https://zenodo.org/record/6839280#.YtFerezMKrM)), rather than the repository and are downloaded when needed.

***
### TODO

- [x] CI caching

***
### Dependencies

#29


.github/workflows/tests.yml
```diff
@@ -46,4 +58,8 @@ jobs:
         shell: bash
         run: |
           export OMP_NUM_THREADS=1
-          py.test ${GITHUB_WORKSPACE}/tdms/tests
+          py.test ${GITHUB_WORKSPACE}/tdms/tests/system/ -s -x
```
t-young31, line 61: `-s` prints stdout and `-x` fails on the first error

.github/workflows/tests.yml
```diff
@@ -21,6 +21,18 @@ jobs:
...
+      - name: Setup cache
+        id: cache-test-data
+        uses: actions/cache@v3
+        with:
+          path: test_data
+          key: test_data
+
+      - name: Copy test data
+        if: steps.cache-test-data.outputs.cache-hit == 'true'
+        run: cp -r test_data/ ${GITHUB_WORKSPACE}/tdms/tests/system/data
```
giordano, line 34: Why not simply caching `${GITHUB_WORKSPACE}/tdms/tests/system/data`?

.gitignore
```diff
@@ -8,3 +8,4 @@ doc/*.pdf
 **/.pytest_cache/
 **.pdf
 **cmake-build**
+tdms/tests/system/data
```
giordano, line 11:
```suggestion
tdms/tests/system/data

```

.github/workflows/tests.yml
t-young31: An excellent question... (blindly copying the example might have had something to do with it)

giordano:
Uh, tests now take some time.  But https://github.com/UCL/TDMS/runs/7424344884?check_suite_focus=true#step:17:2
```
Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
```

PR #35, opened 2022-07-15
Contributing guidelines

### LINKED ISSUE #5: Add contributing guidelines
Adding a set of contributing guidelines (as CONTRIBUTING.md) in the root directory covering: main branch protection, PRs enforced, work on issues, max size of PRs etc.

t-young31:
Closes #5 

Adds a simple set of contributing guidelines – please feel free to push to this branch

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,28 @@
+# Contribution Guidelines
+
+### New features/Issues
+
+To suggest a new feature or report an issue please use the 
+[issue](https://github.com/UCL/TDMS/issues) board providing details of either 
+the feature or the bug with steps to reproduce it. To contribute to the code
+base please (1) [Fork](https://docs.github.com/en/get-started/quickstart/fork-a-repo)
+this repository, (2) Commit and push the changes to your fork and (3) Submit
+a [pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request)
+(PR) against the _main_ branch, being mindful of the guidelines below.
+
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
...
+3. and submit a [pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) (PR) against the _main_ branch, being mindful of the guidelines below.
+
+
+### Pull requests
+
+PRs should follow the following guidelines
+
+- Contain a **context** or **description** of the change
+- Include any useful hints for reviewers
+- Not be excessively large (500 line diff at a maximum)
```
giordano, line 21: This is a good suggestion, but I'm not convinced about putting down a threshold like that.  There may be legitimate cases (for example when refactoring) where you may end up touching more than 500 lines.  How about something along the lines of "keep PRs to a single logically consistent set of changes".  I'm not fond of the wording, but the point is to avoid cramming multiple unrelated changes in a single PR, which I think is the more important principle to follow.

giordano (commented):
Looks overall good to me.  I made some simple changes, and left a comment about a possible rephrasing (but wasn't sure about the specific wording)

CONTRIBUTING.md
t-young31: I mostly agree with this. I guess it depends on the type of refactoring but I'd still be in favour of never having to look at >500 line diffs. Personal preference I guess 🤷🏼 
I'll tone it down appropriately

giordano (approved):

prmunro (approved):

Issue #12, opened 2022-06-13
Add logging


t-young31:
Perhaps [spdlog](https://github.com/gabime/spdlog)? Suggestions welcome!

giordano:
I don't really have much experience with logging in C++.  Out of curiosity, what do you plan to use it for?

t-young31:
There are currently lots of commented out print lines (e.g. [here](https://github.com/UCL/TDMS/blob/2b8d585d9de8fb1974f18cbc201f277a0955ed10/iterater/iterater_OMP_FDTD.cpp#L1182)), that might be nice to convert to logging statements for debugging etc(?)

PR #36, opened 2022-07-21
Openandorder refactor

t-young31:
Refactors `openandorder`. This PR adds:

- A few simple tests of the CLI
- Changes NULL to nullptr (a clangtidy suggestion)
- Extracts some of the main function into functions and argument parser class for readability
- Extracts and argument parser and grid initialiser class


t-young31:
Will be rebased onto #35 when it's merged

tdms/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ project(tdms)
 
 # General setup ---------------------------------------------------------------
 cmake_minimum_required(VERSION 3.21)
+set(CMAKE_CXX_STANDARD 11)
```
t-young31, line 5: can be removed

tdms/src/argument_parser.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+  auto args = ArgumentNamespace(nargs, argv);
+
+  if (args.have_flag("-h")){
+    print_help_message();
+    exit(0);
+  }
+
+  if( (args.num_non_flag != 2) &&  (args.num_non_flag != 3) ){
+    fprintf(stderr,"incorrect number of arguments\n");
```
giordano, line 17: Any specific reason for not using `std::cerr`/`std::cout`?

giordano (commented):
Any way to test the introduced features, like argument parsing?

tdms/tests/system/test_cli.py
```diff
@@ -0,0 +1,17 @@
+import pytest
```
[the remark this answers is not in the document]
t-young31: [yes](https://github.com/UCL/TDMS/blob/8c282314e6b9380a9da7d978171ab586aa6cc366/.github/workflows/tests.yml#L57)!

tdms/src/queuelinked.cpp
```diff
@@ -1,139 +0,0 @@
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
-
-#include "queuelinked.h"
```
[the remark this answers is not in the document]
t-young31: absolutely – thanks for spotting 👍🏼

tdms/src/argument_parser.cpp
t-young31: consistency with the rest of the codebase. I think it'd be nice to switch to configurable logging (as #12) in one fell swoop. happy to modify now if you have strong feelings though

t-young31:
> Any way to test the introduced features, like argument parsing?

At the moment I'd say there are more important things to do than unit testing the argument parsing that has some [system tests](https://github.com/UCL/TDMS/blob/573871944e30d6c18570626017c39c3e9e80bf4a/tdms/tests/system/test_cli.py), but could be persuaded otherwise!

giordano (approved):

prmunro (approved):

willGraham01 (approved):

tdms/include/openandorder.h
```diff
@@ -0,0 +1,12 @@
+#include "matio.h"
```
dstansby, line 1: For consistency, snake case this filename? At a glance I parsed the name as "operand order", so snake case would avoid that and match the naming convention with other files.

tdms/src/argument_parser.cpp
```diff
@@ -0,0 +1,120 @@
...
```
dstansby, line 19:
```suggestion
    fprintf(stderr,"incorrect number of arguments.\n\n");
    print_help_message();
```
It's always helpful to tell the user what to do to fix - I think the help message would work well here for that.

tdms/src/argument_parser.cpp
```diff
@@ -0,0 +1,120 @@
...
+ArgumentNamespace ArgumentParser::parse_args(int nargs, char *argv[]) {
+
+  auto args = ArgumentNamespace(nargs, argv);
+
+  if (args.have_flag("-h")){
+    print_help_message();
+    exit(0);
+  }
+
+  if( (args.num_non_flag != 2) &&  (args.num_non_flag != 3) ){
```
dstansby, line 18:
```suggestion
  if !((args.num_non_flag == 2) || (args.num_non_flag == 3)){
```
I always struggle to get my head around > 1 not statements combined with another operator, and think this is the clearer way to say "If number of args isn't 2 or 3"

tdms/src/argument_parser.cpp
```diff
@@ -0,0 +1,120 @@
...
+            args.have_flag("-m"));
+  }
+
+  return args;
+}
+
+void ArgumentParser::print_help_message(){
+  fprintf(stdout,"Usage:\n"
+                 "openandorder [options] infile outfile\n"
+                 "openandorder [options] infile gridfile outfile\n"
```
dstansby, line 43: Ideally (for simplicity) this would be
```suggestion
                 "openandorder [options] infile outfile gridfile\n"
```
to match the two-arg case. But I guess there's a good reason (backwards compat?) to keep it this way round?

tdms/src/argument_parser.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+const char* ArgumentNamespace::output_filename() {
+
+  if (num_non_flag == 3){
+    return non_flag_arguments[2].c_str();
+  } else if (num_non_flag == 2){
+    return non_flag_arguments[1].c_str();
+  }
+
+  throw std::runtime_error("Failed to determine the output file from arguments");
```
dstansby, line 85: Do you need to bother with this error? Doesn't the constructor ensure `num_non_flag` is 2 or 3?

tdms/src/argument_parser.cpp
```diff
@@ -0,0 +1,120 @@
...
+  throw std::runtime_error("Failed to determine the grid file from arguments");
+}
+
+const char* ArgumentNamespace::input_filename() {
+
+  if (num_non_flag > 0){
+    return non_flag_arguments[0].c_str();
+  }
+
+  throw std::runtime_error("Failed to determine the input file from arguments");
```
dstansby, line 103: Do you need to bother with this error? Doesn't the constructor ensure `num_non_flag` is >0.

tdms/src/argument_parser.cpp
```diff
@@ -0,0 +1,120 @@
...
+  } else if (num_non_flag == 2){
+    return non_flag_arguments[1].c_str();
+  }
+
+  throw std::runtime_error("Failed to determine the output file from arguments");
+}
+
+const char* ArgumentNamespace::grid_filename() {
+
+  if (num_non_flag == 3){
```
dstansby, line 90:
```suggestion
  if has_grid_filename(){
```
Might as well use the method to make intent clear?

tdms/src/utils.cpp
```diff
@@ -0,0 +1,23 @@
...
+  auto file = fopen(filename, mode);
+
+  if(file == nullptr){
+    throw runtime_error("Unable to open file " + string(filename));
+  }
+
+  fclose(file);
+}
+
+bool are_equal(const char* a, const char* b){
```
dstansby, line 21:
```suggestion
bool are_equal_str(const char* a, const char* b){
```
Is a clearer function name?

tdms/tests/system/utils.py
```diff
@@ -132,7 +132,10 @@ def wrapped_function(*args, **kwargs):
     return func_decorator
 
 
-def run_tdms(input_filename: str, output_filename: str) -> None:
+def run_tdms(*args,
+             return_output: bool = False,
+             return_returncode: bool = False
+             ) -> Any:
```
dstansby, line 138:
```suggestion
             ) -> Union[str, bool, None]:
```

tdms/tests/system/utils.py
```diff
@@ -132,7 +132,10 @@ def wrapped_function(*args, **kwargs):
     return func_decorator
 
 
-def run_tdms(input_filename: str, output_filename: str) -> None:
+def run_tdms(*args,
+             return_output: bool = False,
+             return_returncode: bool = False
```
dstansby, line 137: This is confusing if you specify both as `True`. I'd suggest just returning both as a tuple all the time.

dstansby (commented):
I haven't looked at detail into `ftdd_grid_initialiser.cpp`, `matrix_collection.cpp`, or `openandorder.cpp` as they seem copy/pasted and changing the code is outside scope of this PR. Otherwise looks 👍 with some minor suggestions

As a general comment, would be good to get a linter to enforce consisted code style asap. I'm happy to open an issue for that?

dstansby:
Oh, also a rebase onto main to get a CI run in would be good

tdms/src/argument_parser.cpp
t-young31: Thanks for pointing that out, although I actually find that equally difficult to parse. Will create a method 👌🏼

tdms/src/argument_parser.cpp
t-young31: I fear we need some backwards compat, not to break _all_ of Peter's scripts

tdms/src/argument_parser.cpp
t-young31: Definitely (not going to commit your suggestion because it's missing some brackets 😅)

tdms/src/argument_parser.cpp
t-young31: I don't think so(?)

tdms/src/utils.cpp
t-young31: no need for 🦆s here, imo

tdms/tests/system/utils.py
t-young31: totally agree. shouldn't be so sloppy, even in tests!

tdms/src/argument_parser.cpp
dstansby: I thought https://github.com/UCL/TDMS/pull/36/files#diff-10d12f6e6832513c1220f362f528e5a012142d35d32c64f89bfef5cf12190c54R18 ran the check, but maybe I'm missing something.

t-young31: Ah yeah, I thought it conceivable that an argument namespace could be created without that check after?

tdms/tests/system/utils.py
```diff
@@ -132,10 +132,14 @@ def wrapped_function(*args, **kwargs):
     return func_decorator
 
 
-def run_tdms(*args,
-             return_output: bool = False,
-             return_returncode: bool = False
-             ) -> Union[str, bool, None]:
+class Result:
```
dstansby, line 135: This should definitely be a `dataclass`, but that's not a blocker

dstansby (approved):

Issue #14, opened 2022-06-13
Check no e/hfile with illfile


t-young31:
> There should be an error message when a user specifies a particular
efield and hfield file when they also pass an illfile

PR #37, opened 2022-07-22
Add check on field names

### LINKED ISSUE #13: Check illumination file contents
> Error checking where the contents of illfile are loaded, in
> particular, i don't currently check that the loaded file has the
> correct element names, only the number of elements

t-young31:
Closes 
- #13
- #14

### DIFF
--- a/.github/workflows/matlab_tests.yml
+++ b/.github/workflows/matlab_tests.yml
@@ -20,5 +20,5 @@ jobs:
 
       - name: Run script
         uses: matlab-actions/run-command@v1
-        with:
+        with:      # rfpcb contains a delta function definition which must be removed so a delta struct can be defined
           command: rmpath('/usr/local/MATLAB/R2022a/toolbox/rfpcb/rfpcb/shapes'), run('tdms/tests/matlab/run_tests.m')
--- a/.github/workflows/matlab_tests.yml
+++ b/.github/workflows/matlab_tests.yml
@@ -2,11 +2,10 @@ name: MATLAB tests
 
 on:
   push:
-  #push:
-  #  branches: "main"
-  #pull_request:
-  #release:
-  #workflow_dispatch:on: [push]
+    branches: "main"
+  pull_request:
+  release:
+  workflow_dispatch:on: [push]
 
 jobs:
   tests:
--- a/.github/workflows/matlab_tests.yml
+++ b/.github/workflows/matlab_tests.yml
@@ -5,7 +5,7 @@ on:
     branches: "main"
   pull_request:
   release:
-  workflow_dispatch:on: [push]
+  workflow_dispatch:
 
 jobs:
   tests:
--- a/.github/workflows/matlab_tests.yml
+++ b/.github/workflows/matlab_tests.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Set up MATLAB
         uses: matlab-actions/setup-matlab@v1
--- a/.github/workflows/matlab_tests.yml
+++ b/.github/workflows/matlab_tests.yml
@@ -1,0 +1,25 @@
+name: MATLAB tests
+
+on:
+  push:
+  #push:
+  #  branches: "main"
+  #pull_request:
+  #release:
+  #workflow_dispatch:on: [push]
+
+jobs:
+  tests:
+    name: Run MATLAB Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+
+      - name: Set up MATLAB
+        uses: matlab-actions/setup-matlab@v1
+
+      - name: Run script
+        uses: matlab-actions/run-command@v1
+        with:
+          command: rmpath('/usr/local/MATLAB/R2022a/toolbox/rfpcb/rfpcb/shapes'), run('tdms/tests/matlab/run_tests.m')
--- a/tdms/tests/matlab/README.md
+++ b/tdms/tests/matlab/README.md
@@ -1,0 +1,7 @@
+### MATLAB tests
+
+The unit tests in this directory can be run with
+
+```bash
+matlab -nodisplay -nosplash -nodesktop -r "run('run_tests.m');exit;"
+```
--- a/tdms/tests/matlab/README.md
+++ b/tdms/tests/matlab/README.md
@@ -3,5 +3,5 @@
 The unit tests in this directory can be run with
 
 ```bash
-matlab -nodisplay -nosplash -nodesktop -r "run('run_tests.m');exit;"
+matlab -batch run_tests
 ```
--- a/tdms/tests/matlab/data/pstd_input_file.m
+++ b/tdms/tests/matlab/data/pstd_input_file.m
@@ -1,0 +1,141 @@
+%these are not involved in the formal input file spec
+lambda = 1300e-9;
+
+%size of Yee cell in metres
+delta.x = lambda/4;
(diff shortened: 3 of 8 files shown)


t-young31:
File used to test the function (based on examples/arc_01)

```matlab
addpath('../../tdms/matlab');
[x,y,z,lambda] = fdtd_bounds('pstd_input_file.m');

rad = 15e-6;
%refractive index of cylinder
refind = 1.42;
y = 0;
[X,Y,Z] = ndgrid(x,y,z);

I = zeros(size(X));
I( find( (X.^2 + Z.^2) < rad^2 )) = 1;
I( (end-3):end,1,:) = 0;
I( :, 1, (end-3):end) = 0;
inds = find(I(:));
[ii,jj,kk] = ind2sub(size(I), inds);
composition_matrix = [ii jj kk ones(size(ii))];
material_matrix = [1 refind^2 1 0 0 0     0     0     0 0 0];
save(sprintf('gridfiles/gridfile_cyl'), 'composition_matrix', 'material_matrix');

Isource = zeros(size(X));
Jsource = zeros(size(X));
Ksource = zeros(size(X));

save(sprintf('tmp'), 'Isource', 'Jsource', 'Jsource');

[fdtdgrid, Exs,Eys,Ezs,Hxs,Hys,Hzs,grid_labels,camplitudes,vertices,facets,Id] = ...
iteratefdtd_matrix('pstd_input_file.m','filesetup','in/pstd_cyl','gridfiles/gridfile_cyl.mat','tmp.mat');
```

giordano (approved):
Looks good to me, as far as I can tell (not super familiar with matlab though!).  My understanding is that the new check throws an error if the condition is not met, while previously the code will continue running, is the new behaviour intended?

t-young31:
> is the new behaviour intended?

It's the behaviour I intended to write.. @prmunro is perhaps better placed to answer whether this is the intended outcome. Silently ignoring something is worse than throwing an error, in my opinion

prmunro:
The above test won't result in the modified code being executed. I am in the process of generating a test that will.

t-young31:
> The above test won't result in the modified code being executed. I am in the process of generating a test that will.

Indeed the previous script didn't check the exi/eyi could be present in the file. I've now added some specific unit tests which are run on Github actions. [Results from the last run](https://github.com/UCL/TDMS/runs/7558560046?check_suite_focus=true#step:4:9)

.github/workflows/matlab_tests.yml
```diff
@@ -0,0 +1,24 @@
...
+  release:
+  workflow_dispatch:
+
+jobs:
+  tests:
+    name: Run MATLAB Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
```
dstansby, line 16:
```suggestion
        uses: actions/checkout@v3
```

tdms/tests/matlab/README.md
```diff
@@ -0,0 +1,7 @@
+### MATLAB tests
+
+The unit tests in this directory can be run with
+
+```bash
+matlab -nodisplay -nosplash -nodesktop -r "run('run_tests.m');exit;"
```
dstansby, line 6:
```suggestion
matlab -batch run_tests
```
This is how `MATLAB` recommends running a script without launching a window. I can't find the specific documentation right now I'm afraid (MATLAB documentaiton is a maze...)

tdms/tests/matlab/data/pstd_input_file.m
```diff
@@ -0,0 +1,141 @@
+%these are not involved in the formal input file spec
```
dstansby, line 1: Add a quick comment at the top of this file that explains what it is? Something like "Sample input file used for running unit tests" or similar I imagine.

tdms/tests/matlab/run_tests.m
```diff
@@ -0,0 +1,15 @@
...
+import matlab.unittest.selectors.HasBaseFolder;
+
+addpath(genpath('.'));
+
+suite = testsuite(pwd, 'IncludeSubfolders', true);
+suite = suite.selectIf(HasBaseFolder(ContainsSubstring('tests')));
+
+runner = TestRunner.withTextOutput('OutputDetail', Verbosity.Detailed);
+results = runner.run(suite);
+
```
dstansby, line 13: If you wnat to add code coverage (here or another PR):

https://github.com/dstansby/matlab-sandbox/blob/6da492672f240df162810cf9948b4a2bc8377841/run_tests.m#L22

tdms/tests/matlab/test_iteratefdtd_matrix.m
```diff
@@ -0,0 +1,55 @@
+
+function tests = test_iteratefdtd_matrix_function
+tests = functiontests(localfunctions);
```
dstansby, line 3:
```suggestion
    tests = functiontests(localfunctions);
```

tdms/tests/matlab/test_iteratefdtd_matrix.m
```diff
@@ -0,0 +1,55 @@
...
+end
+
+function testInvalidIlluminationSourceIJKErrors(testCase)
+    verifyError(testCase, @()createInputWithInvalidIlluminationSourceIJ(), 'TDMSException:InvalidIlluminationFile');
+    delete *.mat
```
dstansby, line 8: Probably better to create and then delete a specific directory?

tdms/tests/matlab/test_iteratefdtd_matrix.m
```diff
@@ -0,0 +1,55 @@
...
+    addPathAndCreateGridFile();
+    exi = zeros(size(1));
+    % Illumnation file must both exi and eyi
+    save(sprintf('invalid_illumnation_file'), 'exi', 'exi');
+
+    [~] = iteratefdtd_matrix('pstd_input_file.m','filesetup',...
+    'tmp_input','gridfile_cyl.mat','invalid_illumnation_file.mat');
+end
+
+function addPathAndCreateGridFile()
```
dstansby, line 39: I think this deserves a quick comment explaining what it does at the top

dstansby (commented):
Overall looks 👍 I've added some suggetions - enough for me to hold off approving for now.

tdms/tests/matlab/README.md
t-young31: love it 👍🏼

tdms/tests/matlab/run_tests.m
t-young31: At this state I'm not too worries about adding codecov, but would definitely be nice in the future

tdms/tests/matlab/test_iteratefdtd_matrix.m
t-young31: Ah yes – the right way. Done now 🎉

dstansby (approved):

prmunro:
[arc_04.zip](https://github.com/UCL/TDMS/files/9212413/arc_04.zip)

Here is a test (using top level file run_pstd_bscan.m) that shows when an error should and should not be generated. I agree that an error should be generated if an incorrect Ksource, Jsource or Isource.

prmunro:
[arc_05.zip](https://github.com/UCL/TDMS/files/9219113/arc_05.zip)
Here is another test that shows correct usage of the data stucture with members exi and eyi, and one case where usage is incorrect and should generate an error.

t-young31:
> Here is a test (using top level file run_pstd_bscan.m) that shows when an error should and should not be generated.

Many thanks for those – they've now been incorporated into the unit test suite

*** 
> Are these tests useful, or are they an overkill?

More tests are always useful (in my opinion at least). I'll add those in and re-request your review when it's done – thanks

prmunro (approved):

Issue #41, opened 2022-07-22
[CI] Build TDMS with `-Wall -Werror` and run a linter


giordano:
For better future-proof development of this package, I think we should

1. clear up all warnings.  In particular, we should
    1. [x] i. build the package in CI with `-Wall`
    2. [ ] ii. once all warnings are solved, build in CI with `-Werror`, to make sure no more warnings are introduced.  To be clear: I'm ***not*** suggesting to use `-Wall -Werror` by default for all users, that's a disruptive setup
2. use a linter, like [`clang-tidy`](https://clang.llvm.org/extra/clang-tidy/) in CI:
    1. [ ] i. I think we can start with a job which runs `clang-tidy` and reports annotations if there are warnings in the changed lines
    2. [ ] ii. Ideally, `clang-tidy` should always come back clean, and if it doesn't the job should fail.  But this can be enabled at a later stage, when all problems are resolved

Note that point 1.ii depends on 1.i, and point 2.ii depends on 2.i, but point 2.i is independent from whole of point 1 (although warnings reported by `clang-tidy` will usually be a superset of those reported by `-Wall`).

https://github.com/cpp-linter/cpp-linter-action looks a decent linter action, for point 2.  I don't think 2 (and in particular 2.ii) is a high priority (also because `clang-tidy` can be particularly over-zealous), but in the long term it may be helpful.

t-young31:
possible iii. use [`clang-format`](https://clang.llvm.org/docs/ClangFormat.html) to enforce a consistent C++ code style

samcunliffe:
Automatically closed with merge of #222.
Reopening because we still want `cpplint` (and maaaybe `clang-tidy`). Discussion ongoing and in the comments of #222.

giordano:
Looking at the issue description above, I think we're already building with `-Werror` https://github.com/UCL/TDMS/blob/342f9a11ff29ffc92c04626826f3b956b882710d/.github/workflows/ci.yml#L103 which is point 1.ii, but not with `-Wall`, which is point 1.i, which is also checked (incorrectly?).  It may be worth giving `-Wall` a try on the same line.

samcunliffe:
[Good old MacOS](https://github.com/UCL/TDMS/actions/runs/4829408578/jobs/8604424875)

samcunliffe:
Seems we won't achieve this in the ARC project. Closing for now.

PR #42, opened 2022-07-22
Compile TDMS with `-Wall`

giordano:
This implements point 1.i in #41.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,6 +44,8 @@ jobs:
         working-directory: ${{runner.workspace}}/build
         run: |
           cmake ${GITHUB_WORKSPACE}/tdms -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
+        env:
+          CXXFLAGS: -Wall
       - name: Build TDMS
         working-directory: ${{runner.workspace}}/build
         shell: bash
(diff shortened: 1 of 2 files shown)


giordano:
Many new warnings, for example https://github.com/UCL/TDMS/runs/7476420841?check_suite_focus=true#step:8:47:
```
/home/runner/work/TDMS/TDMS/tdms/src/tdms_iterator.cpp:3399:10: warning: ‘Enp1’ may be used uninitialized in this function [-Wmaybe-uninitialized]
 3399 |     Enp1 += Cb*dy*Jcxy[k][j][i];
      |     ~~~~~^~~~~~~~~~~~~~~~~~~~~~
```
There are a hundred of similar possibly uninitialised variables.

.github/workflows/tests.yml
```diff
@@ -44,6 +44,8 @@ jobs:
         working-directory: ${{runner.workspace}}/build
         run: |
           cmake ${GITHUB_WORKSPACE}/tdms -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
+        env:
+          CXXFLAGS: -Wall
```
t-young31, line 48: I wonder if this would be better in CMakeLists.txt, so local compilation also throws those warnings. Definitely agree that `-Werror` should only be used in CI

giordano: Fair enough, I changed it to always build with `-Wall` :+1:

t-young31 (approved):

Issue #43, opened 2022-07-25
Test compilation for Windows


t-young31:
TDMS has only been built on MacOS and linux thus far. We need to check this is also possible on Windows

giordano:
Reading the docs of [`matlab-actions/setup-matlab`](https://github.com/matlab-actions/setup-matlab) it wasn't clear to me whether it supports Windows (or macOS), maybe we can just try to enable those platforms and see what happens, maybe it just works?

t-young31:
Lets give it a go! [This comment](https://github.com/matlab-actions/setup-matlab/issues/18#issuecomment-1006990188) suggests it might be possible

PR #48, opened 2022-07-25
[CI] Run on Windows

giordano:
Test in view of #43, I don't know if this will work, let's see

### DIFF
--- a/.github/workflows/linux_tests.yml
+++ b/.github/workflows/linux_tests.yml
@@ -1,4 +1,4 @@
-name: Run tests
+name: Linux tests
 
 on:
   push:
@@ -33,19 +33,10 @@ jobs:
       - name: Set up MATLAB
         uses: matlab-actions/setup-matlab@v1
 
-      - name: Install dependencies Linux
-        if: runner.os == 'Linux'
+      - name: Install dependencies
         run: |
           sudo apt-get update
           sudo apt-get install libfftw3-dev libgomp1 python3
-
-
-      - name: Install dependencies Windows
-        if: runner.os == 'Windows'
-        shell: pwsh
-        run: |
-          conda install fftw --yes
-          echo "FFTWDIR=C:\Miniconda\envs\test\Library" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append
 
       - name: Install Python dependencies
         shell: bash
--- a/.github/workflows/linux_tests.yml
+++ b/.github/workflows/linux_tests.yml
@@ -17,6 +17,7 @@ jobs:
         os:
           - ubuntu-latest
         build_type:
+          - Test
           - Release
     steps:
       - name: Check out repository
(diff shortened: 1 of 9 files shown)


giordano:
Some progress: compilation is successful on Windows, but tests are failing because `tdms.exe` can't be found.

giordano:
I'm completely stuck at this point and I have no clue of what to do.  There are two orthogonal issues:

1. `tdms.exe` is built successfully but it isn't installed.  This could be solved by running `cmake --install . --config ${{ matrix.build_type }}`, but this would install the program in a directory (I think it was something like `C:\Program Files (x86)\tdms\bin`) which isn't in the `PATH` and so the executable can't be found anyway, which isn't much progress.  I see two options:
    1. Search the executable also in a `Release/` subdirectory of the current folder, but I don't like this option much
    2. Add the installation directory to the `PATH`.  I'd say this is the "proper" solution in the Unix world, but I'm not sure people on Windows are accustomed to fiddling with these settings
2. For the life of me I can't make the dynamic loader on Windows find the needed dynamic libraries, like `fftw3.dll` or `libmat.dll`/`libmex.dll`.  I tried to add the relevant directories to `PATH` (the Windows equivalent of `LD_LIBRARY_PATH`), to no avail.  The only working solution is to...`cd` to directory where `fftw3.dll` is, but we definitely can't expect users to move around (and I also have opinions, not positive ones, about this behaviour of the loader).  Also, we can't expect fftw _and_ matlab libraries to be in the same directory

t-young31:
I've also just spent an hour or two on this and I am also stuck.

1. I think moving the executable from build/Release to build/ would be a reasonable solution to combat the hilariously inconsistent behaviour
2. Is more troubling.. I also found I needed to copy the `.dll`s into the same directory 😕 , which is not very user friendly. There also doesn't seem to be a way of using any kind of RPATH so I'm also unsure of the solution.

I suppose a static link to fftw and no matlab dependency would solve this, depending on our priorities we may want to leave this for a bit

prmunro:
I'm not sure whether my review is required here or not. Are you asking whether to persist with this? I don't think I can offer a solution to this. I only ever compiled it under windows using Visual Studio. I understand why you're using cmake though.

giordano:
Note that we aren't having difficulties with compiling the program, but with running it: it can't find fftw and matlab shared libraries

prmunro (approved):
I'm happy to approve this noting that at present the code won't run under windows. Can we add this as something to come back to?

t-young31:
TDMS is now successfully building on GitHub actions under Windows! Turns out setting $PATH can work to find runtime .dlls (and is required because the MATLAB shared libraries have dependencies beyond just libmat.dll libmex.dll and libmx.dll). Due to some oddness with the zip file handling not all the system tests are run, but have instead been checked using an interactive session (following the CI build config/compile). Given this now closes #43 I'd be happy to merge this in and add an issue for running a useful system test on Windows CI, if required.

willGraham01 (approved):

prmunro (approved):

Issue #54, opened 2022-07-27
Demonstration code: MATLAB `save()` errors


willGraham01:
The demonstration script `run_pstd_bscan.m` throws a MATLAB error when run.

MATLAB attempts to save the generated input/output files (that are to be passed to tdms) to the sub-directories `gridfiles/`, `in/`, and `out/`, all of which don't exist after cloning the repository (and successfully installing tdms) which causes the `save()` function to error.

MATLAB needs to check whether these directories need to be created, then call to `mkdir` if necessary, prior to attempting to generate the input and output files.

PR #57, opened 2022-08-01
#54 run_pstd_bscan.m now creates missing directories

willGraham01:
#54 Fixes error thrown in `run_pstd_bscan.m` by getting MATLAB to create the directories `gridfiles/`, `in/`, and `out/`, if they do not exist prior to running the script.

Also some other things I'm dumping here as afterthoughts:
- Rather than displaying the figures in MATLAB, we could just save them as images? This would mean the MATLAB GUI wouldn't need to load up every time, and we can stick to something like `matlab -nodesktop -nosplash -r run_pstd_bscan` in the command line.
- On a note related to the above, MATLAB does not exit upon completion of a script, unless you explicitly give it the `exit()` command at the end of the script!

### DIFF
--- a/examples/arc_01/run_pstd_bscan.m
+++ b/examples/arc_01/run_pstd_bscan.m
@@ -31,7 +31,7 @@ save('gridfile_cyl', 'composition_matrix', 'material_matrix');
 composition_matrix = [];
 save(sprintf('gridfile_fs'), 'composition_matrix', 'material_matrix');
 
-%generate C input files and save to directory in/
+%generate tdms executable input files
 [fdtdgrid, Exs,Eys,Ezs,Hxs,Hys,Hzs,grid_labels,camplitudes,vertices,facets,Id] = iteratefdtd_matrix('pstd_input_file.m','filesetup','in_pstd_cyl','gridfile_cyl.mat','');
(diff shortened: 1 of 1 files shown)


t-young31:
Looks great – thanks for fixing this. I wonder if it'd be better to just rename the input and output files and not save anything in directories? I'd maybe try to keep the example/s as small as possible to emphasise the TDMS function calls. Happy to approve this and merge it in for the time being and add an issue to make some clearer example in the future? 

*** 

> Rather than displaying the figures in MATLAB, we could just save them as images?

Good shout. More example are definitely better, so maybe we could add an example (e.g. one of _fs or _cyl) that could be run without the GUI

willGraham01:
> I wonder if it'd be better to just rename the input and output files and not save anything in directories? I'd maybe try to keep the example/s as small as possible to emphasise the TDMS function calls
Almost definitely - the (one!) example is already in its own subdirectory within `examples/`, so I agree that extra directories are kind of redundant. I'll quickly change this to avoid making even more directories.

> Good shout. More example are definitely better, so maybe we could add an example (e.g. one of _fs or _cyl) that could be run without the GUI
I'll make another issue for this (and more examples in general!). We don't have any other examples of course, but I can edit `arc-01` to be run without the GUI and save the output to figures, and pre emptively call it `arc-02`?

PR #59, opened 2022-08-02
Add linting for Python code

dstansby:
Now Python is being introduced to run tests, this PR introduces config and a GH actions job to enforce linting on the codebase. This still needs some developer documentation adding, so opening as a draft for now.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,0 +1,17 @@
+name: Run code linting
+
+on:
+  push:
+    branches: "main"
+  pull_request:
+  release:
+  workflow_dispatch:
+
+jobs:
+  pre-commit:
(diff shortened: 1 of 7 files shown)


dstansby:
👍 I've added a small bit to the developer docs on applying the style guide to Python files. Suggestions very welcome if it's too short or unclear (I'm too used to using `pre-commit` now!)

.github/workflows/linting.yml
```diff
@@ -0,0 +1,17 @@
+name: Run code linting
```
samcunliffe, line 1:
```suggestion
name: Run python code linting
```
I would specify python code linting (either that, or add some C++ linter).

samcunliffe (approved):

Issue #44, opened 2022-07-25
Modify interpolation scheme


t-young31:
Overhaul interpolation from the split grid to the central grid. 

1. Different scheme required for each field component.
2. Is implemented for the electric field, although interpolateTimeDomainFieldCentralEBandLimited was hacked to work for the 2D case, this should be checked.
3. More significant work required for the magnetic field
4. Enable interpolation to be performed right up to the edge of the computational domain.

***
### Dependencies:

- [x] Theoretical outline from @prmunro 

***
### Tasks specified in `PSTD_interpolation_scheme.pdf`:

- [x] Write (time-domain) interpolation methods for E-field components. Should use BLi, and if not possible, fall back on cubic interpolation. Closed by #62.
- [x] Write (time-domain) interpolation methods for H-field components using the 2D-method interpolation method. Interpolation in each dimension should again use BLi if possible, and failing that fall back on cubic. Closed by #62.
- [x] #112
- [x] Unit tests for cubic interpolation methods. Closed by #62.
- [x] #111
- [x] #110
- [ ] #109
- [x] #113
- [ ] #114
- [x] #115

giordano:
Related to interpolation, there are many interpolation functions which are supposed to return `int`, but the return value is never actually used.  For example https://github.com/UCL/TDMS/blob/805e5dd601168ea1bfb16e32d3dd326c6a397302/tdms/src/interpolate.cpp#L1086 can return `-1` under certain conditions https://github.com/UCL/TDMS/blob/805e5dd601168ea1bfb16e32d3dd326c6a397302/tdms/src/interpolate.cpp#L1094-L1096 but its use at https://github.com/UCL/TDMS/blob/805e5dd601168ea1bfb16e32d3dd326c6a397302/tdms/src/tdms_iterator.cpp#L6174-L6176 doesn't check the return value, so that an error would never be caught.

I guess these functions should either be `void` and throw errors when needed, or at very least the return values should be checked and appropriately dealt with.

willGraham01:
Adding another task list to this issue in light of the information we have from the shared document `PSTD_interpolation_scheme.pdf`.

- [x] Write (time-domain) interpolation methods for E-field components. Should use BLi, and if not possible, fall back on cubic interpolation. Closed by #62.
- [ ] Write (time-domain) interpolation methods for H-field components using the 2D-method interpolation method. Interpolation in each dimension should again use BLi if possible, and failing that fall back on cubic.
[...]

PR #62, opened 2022-08-08
Band Limited Interpolation Functionality

willGraham01:
Addresses the following tasks as laid out in [Band limited interpolation scheme for
PSTD/FDTD method](https://www.overleaf.com/project/62e84bb1f1395dea4e8f7e8d), and also #44.
Requires the following work:

- [x] Band Limited interpolation (BLi) for the E field only considers when there are 4 points available either side, if this is not the case it should fall back onto cubic interpolation.
- [ ] BLi for the H field, using interpolation in two dimensions, needs to be implemented
- [ ] `fieldsample` does not allow for extraction of magnetic field at the vertices - this must be added.
- [ ] Write generic functions for the interpolation of the E- and H-fields in both the time domain (real data) and and frequency domain (complex data)
- [ ] Call the above functions in `exphasorssurface`, `exphasorsvolume`, and `fieldsample`


tdms/tests/unit/test_interpolate.cpp
```diff
@@ -0,0 +1,128 @@
...
+TEST_CASE("checkInterpolationPoints: exceptions thrown") {
+
+    // setup some fake field dimensions
+    int I = 6, J = 7, K = 8;
+
+    // 2D simulations have J = 0 - no width in this dimension
+    // checkInterpolationPoints should not throw an exception if we provide j_l = 2 > j_u = -2, as is done for a 2D simulation (J=0), and request interpolation over the maximum i,k range
+    CHECK_NOTHROW(checkInterpolationPoints(2, I - 2, 2, -2, 2, K - 2, I, 0, K));
+
+    // conversely, it should throw an error if we try to provide j_l = 0 = j_u, which one may percieve are sensible choices, in this case
```
t-young31, line 20: > which one may perceive are sensible choices 

That's exactly what I would perceive. Is the j_l = j_u != 0 requirement from the previous implementation?

tdms/tests/unit/test_interpolate.cpp
```diff
@@ -0,0 +1,128 @@
...
+    - j_u > J-2 with text Interpolation error: j_u too large
+    - k_u > K-2 with text Interpolation error: k_u too large */
+    CHECK_THROWS_AS(checkInterpolationPoints(1, I - 2, 2, J - 2, 2, K - 2, I, J, K), runtime_error);
+    CHECK_THROWS_AS(checkInterpolationPoints(2, I - 1, 2, J - 2, 2, K - 2, I, J, K), runtime_error);
+    CHECK_THROWS_AS(checkInterpolationPoints(2, I - 2, 0, J - 2, 2, K - 2, I, J, K), runtime_error);
+    CHECK_THROWS_AS(checkInterpolationPoints(2, I - 2, 2, J - 0, 2, K - 2, I, J, K), runtime_error);
+    CHECK_THROWS_AS(checkInterpolationPoints(2, I - 1, 2, J - 2, -1, K - 2, I, J, K), runtime_error);
+    CHECK_THROWS_AS(checkInterpolationPoints(2, I - 1, 2, J - 2, 2, K + 1, I, J, K), runtime_error);
+
+    /* When provided input arguments that do not satisfy any of the above - even cases such as i_l > i_u - should pass without throwing an exception */
```
t-young31, line 38: could maybe separate these out into another test with a more informative name

tdms/tests/unit/test_interpolate.cpp
```diff
@@ -0,0 +1,128 @@
...
+ * @brief In the case when cubic interpolation is to be used, check that all polynomial fields up to cubic order are interpolated exactly (to within machine error)
+ * 
+ */
+TEST_CASE("interp: cubic interpolation is exact") {
+
+    // equidistant points
+    double x[] = {0.,1.,2.,3.};
+    // test acceptence tolerance. Allow for FLOP inprecision and rounding errors
+    // error should be \approx 4 max(c_i) x^2 __DBL__EPSILON, so order 40 * __DBL__EPSILON__
+    double tol = 4e1 * __DBL_EPSILON__;
```
t-young31, line 87: noice

tdms/tests/unit/test_interpolate.cpp
```diff
@@ -0,0 +1,128 @@
...
+    double c3 = 4.2;
+    v1 += c3*x[0]*x[0]*x[0]; v2 += c3*x[1]*x[1]*x[1]; v3 += c3*x[2]*x[2]*x[2]; v4 += c3*x[3]*x[3]*x[3];
+    v12 += c3 * (x[1] + x[0]) * (x[1] + x[0]) * (x[1] + x[0]) / 8.;
+    v23 += c3 * (x[2] + x[1]) * (x[2] + x[1]) * (x[2] + x[1]) / 8.;
+    v34 += c3 * (x[3] + x[2]) * (x[3] + x[2]) * (x[3] + x[2]) / 8.;
+
+    CHECK(abs(v12 - interp2(v1, v2, v3, v4)) <= tol);
+    CHECK(abs(v23 - interp1(v1, v2, v3, v4)) <= tol);
+    CHECK(abs(v34 - interp3(v1, v2, v3, v4)) <= tol);
+}
```
t-young31, line 128: mose _really_ likes new lines at the end of files 😅

tdms/src/interpolate.cpp
```diff
@@ -67,71 +103,48 @@ void interpolateFieldCentralE( double ***Ex_yee, double ***Ey_yee, double ***Ez_
...
+ * @param[in] J Number of elements in the j direction of the FDTD grid
+ * @param[in] K Number of elements in the k direction of the FDTD grid
+ * @param[in] i_l Least i index into the FDTD grid to evaluate the field at. Should be >= 2
+ * @param[in] i_u Greatest i index into the FDTD grid to evaluate the field at. Should be <= I-2
+ * @param[in] j_l Least j index into the FDTD grid to evaluate the field at. Should be >= 2
+ * @param[in] j_u Greatest j index into the FDTD grid to evaluate the field at. Should be <= J-2
+ * @param[in] k_l Least k index into the FDTD grid to evaluate the field at. Should be >= 2
+ * @param[in] k_u Greatest k index into the FDTD grid to evaluate the field at. Should be <= K-2
  */
 void interpolateFieldCentralE_TE( double ***Ex_yee, double ***Ey_yee, double ***Ez_yee,
```
t-young31, line 149: these can now just be deleted?!

tdms/src/interpolate.cpp
```diff
@@ -204,76 +192,53 @@ void interpolateFieldCentralE_TM( double ***Ex_yee, double ***Ey_yee, double ***
...
-    
+    throw runtime_error("Error in mxInterpolateFieldCentralE, Ex_yee does not have 3 dimensions\n");   
   }
   //fprintf(stderr, "mxInterpolateFieldCentralE Pos 01\n");
+
   indims = (int *)mxGetDimensions( (mxArray *)Ex_yee);
   //fprintf(stderr, "mxInterpolateFieldCentralE(indims): (%d,%d,%d)\n",indims[0],indims[1],indims[2]);
   //fprintf(stderr, "mxInterpolateFieldCentralE: j_u: %d, j_l: %d\n",j_u,j_l);
-  //assume that all matrices have the same dimensions
+  //assume that all matrices have the same dimensions - DANGEROUS, SHOULDN'T WE CHECK THIS??
```
t-young31, line 241: > DANGEROUS, SHOULDN'T WE CHECK THIS??

I guess it depends how performance critical this function is.. my guess it not very as it exists outside of the main loop(?) in which case lets definitely add a check!

tdms/src/interpolate.cpp
```diff
@@ -695,90 +629,62 @@ void interpolateFieldCentralH_TE( double ***Hx_yee, double ***Hy_yee, double ***
...
- *J[in] - number of elements in the j direction of the FDTD grid
- *K[in] - number of elements in the k direction of the FDTD grid
- *
- *i_l[in]  - lowest i index into the FDTD grid to evaluate the field at. Should be >= 2.
- *i_u[out] - largest i index into the FDTD grid to evaluate the field at. Should be <= I-2.
- *j_l[in]  - lowest i index into the FDTD grid to evaluate the field at. Should be >= 2.
- *j_u[out] - largest i index into the FDTD grid to evaluate the field at. Should be <= J-2.
- *k_l[in]  - lowest i index into the FDTD grid to evaluate the field at. Should be >= 2.
- *k_u[out] - largest i index into the FDTD grid to evaluate the field at. Should be <= K-2.
+ * @param[in] Hx_yee Steady state x component of magnetic field calculated at points in the Yee cell
```
t-young31, line 673: we definitely need a struct for these – so. many. params.

tdms/src/interpolate.cpp
```diff
@@ -1393,3 +1272,177 @@ void interpolateTimeDomainFieldCentralE_TM(  double ***Exy, double ***Exz, doubl
...
+/**
+ * @brief Determines the appropriate interpolation scheme to use
+ * 
+ * @param[in] cells_in_direction The number of Yee cells in the interpolation direction of interest
+ * @param[in] cell_id The current ID (in this dimension) of the Yee cell
+ * @return int Equal to BAND_LIMITED, INTERP1, INTERP2, INTERP3, indicating the appropriate interpolation scheme
+ */
+int determineInterpScheme(int cells_in_direction, int cell_id) {
+  // interpolation is impossible if the total number of cells in this direction is <4
+  if (cells_in_direction <= 4) { throw out_of_range("Error: computational domain has fewer than 4 cells in at least 1 dimension, cubic and band-limited interpolation impossible.\n"); }
```
t-young31, line 1285: from a code style point of view I'd prefer if these `throw` s were on the next line. but we are definitely going to need to clang-format everything anyway

tdms/src/interpolate.cpp
```diff
@@ -1393,3 +1272,177 @@ void interpolateTimeDomainFieldCentralE_TM(  double ***Exy, double ***Exz, doubl
...
 }
+
+/**
+ * @brief Determines the appropriate interpolation scheme to use
+ * 
+ * @param[in] cells_in_direction The number of Yee cells in the interpolation direction of interest
+ * @param[in] cell_id The current ID (in this dimension) of the Yee cell
+ * @return int Equal to BAND_LIMITED, INTERP1, INTERP2, INTERP3, indicating the appropriate interpolation scheme
+ */
+int determineInterpScheme(int cells_in_direction, int cell_id) {
```
t-young31, line 1283:
```suggestion
InterpolationScheme determineInterpScheme(int cells_in_direction, int cell_id) {
```
where `InterpolationScheme` is an enum? may be slightly more readable

tdms/src/interpolate.cpp
```diff
@@ -1393,3 +1272,177 @@ void interpolateTimeDomainFieldCentralE_TM(  double ***Exy, double ***Exz, doubl
...
+  // interpolation is impossible if the total number of cells in this direction is <4
+  if (cells_in_direction <= 4) { throw out_of_range("Error: computational domain has fewer than 4 cells in at least 1 dimension, cubic and band-limited interpolation impossible.\n"); }
+  // Yee cell with index <=0 doesn't exist (indexing starts from 1)
+  else if (cell_id <= 0) { throw out_of_range("Error: Interior Yee cell index <=0 requested (must be >=1).\n");}
+  // Yee cell with index >=cells_in_direction doesn't exist
+  else if (cell_id > cells_in_direction) { throw out_of_range("Error: requested Yee cell index beyond maximum number of Yee cells.\n");}
```
t-young31, line 1289:
```suggestion
  else if (cell_id >= cells_in_direction) { throw out_of_range("Error: requested Yee cell index beyond maximum number of Yee cells.\n");}
```

tdms/src/interpolate.cpp
t-young31: or, maybe even better [return a function](https://stackoverflow.com/questions/31387238/c-function-returning-function) (provided the signatures are the same)?

tdms/src/interpolate.cpp
```diff
@@ -1393,3 +1272,177 @@ void interpolateTimeDomainFieldCentralE_TM(  double ***Exy, double ***Exz, doubl
...
+ * We enforce {a,b,c} = {i,j,k}, with the conventions a=i, b=j, c=k; a=j, b=i, c=k; a=k, b=i, c=j.
+ * Indices and variables are named assuming the a=i case, EG (a,b,c) will specify the index of the Yee cell.
+ * 
+ * @param[in] aID Identifies which direction i(0), j(1), or k(2) the a-direction represents
+ * @param[in] Eab, Eac Split components of the Yee cell
+ * @param[in] a,b,c Index of the Yee cell
+ * @param[in] A Total number of Yee cells in the a-direction
+ * @param[out] Ea The interpolated field component at the centre of the Yee cell
+ */
+void interpolateTimeDomainEcomponent(int aID, double ***Eab, double ***Eac, int a, int b, int c, int A, double *Ea)
```
t-young31, line 1323: not looked at this yet! need me to understand some more background first 😅

t-young31 (commented):
looks great 🌮

tdms/src/interpolate.cpp
willGraham01: You're going to _love_ what I've got so far for the 2D H-field interpolation...

tdms/include/interpolate.h
```diff
@@ -1,6 +1,9 @@
+# include "matrix.h"
+
 double interp1(double v1, double v2, double v3, double v4);
 double interp2(double v1, double v2, double v3, double v4);
 double interp3(double v1, double v2, double v3, double v4);
+void checkInterpolationPoints(int i_l, int i_u, int j_l, int j_u, int k_l, int k_u, int I, int J, int K);
```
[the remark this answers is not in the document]
willGraham01: No worries - I was going with the original style in `interpolate.cpp` which is currently camelCase. Happy to change to whichever we decide later

tdms/tests/unit/test_interpolate.cpp
```diff
@@ -0,0 +1,128 @@
+# include "catch2/catch_test_macros.hpp"
+# include "interpolate.h"
+
+using namespace std;
+
+/**
+ * @brief Test whether checkInterpolationPoints throws exceptions in the correct cases.
```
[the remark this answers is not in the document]
willGraham01: Been keeping to [this one](https://developer.lsst.io/cpp/api-docs.html), only in lieu of it being VS Code's default...

tdms/tests/unit/test_interpolate.cpp
willGraham01: Yep - original code seems to rely on this behaviour, so I thought it best to preserve it for now

t-young31: @prmunro : is the correct behaviour? it seems a little odd

tdms/include/interpolate.h
```diff
@@ -73,3 +76,17 @@ void interpolateTimeDomainFieldCentralH_TM( double ***Hxy, double ***Hxz, double
...
 					int i, int j, int k,
 						   double *Hx, double *Hy, double *Hz);
+
+// define aliases for interpolation scheme flags, for readability
+#define BAND_LIMITED 0
+#define INTERP1 1
+#define INTERP2 2
+#define INTERP3 3
+// aliases for the dimension "a" represents when interpolating a single field component
+#define ax 0
```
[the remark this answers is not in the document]
willGraham01: Will restyle like this

tdms/src/interpolate.cpp
t-young31: maybe the comment is wrong but the code was correct? if it's 1 indexed. sorry for the perhaps poor suggestion!

tdms/src/interpolate.cpp
willGraham01: They're still called (for the time being) in `iterator.cpp`, but they can definitely go once I've tested `interpolatetimeDomainEcomponent` is doing what it should be

tdms/tests/unit/test_interpolate.cpp
willGraham01: Have taken the "no exceptions thrown" checks out into their own test_case

tdms/src/interpolate.cpp
willGraham01: No I think your original comment is correct - the documentation says that Yee cell IDs start from 1,1,1, BUT C indexes start at 0,0,0. It's also compounded by the fact that we _can_ interpolate when we hit the last cell, but can't when we hit the first, since Yee cells are associated to field values at gridpoints _after_ their origin. 
So `cell_id = 0` errors since we would need the field value in the "-1"th Yee cell, whilst `cell_id = cells_in_direction` errors due to memory access faults.
`cell_id = cells_in_direction-1` _doesn't_ error since we can use `interp3`, as there is a field value after this point.

tdms/src/interpolate.cpp
willGraham01: Have turned this into an enum. There's scope to turn it into a function pointer but band-limited interpolation takes a different number of arguments to the `interp` functions, and function pointers are one of my least favourite things. Defo something that can be done once I've got band-limited working though

tdms/include/interpolate.h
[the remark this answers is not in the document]
t-young31: cool – there doesn't seem to be an awful lot of consistency e.g. [is_conductive](https://github.com/UCL/TDMS/blob/af3fa4cba628607cfa8d6b65e34e8fd574c601cf/tdms/src/iterator.cpp#L7184) so.. 🤷🏼  
lets enforce 🐍case from here?

tdms/tests/unit/test_interpolate.cpp
prmunro: @t-young31 , can you please direct me to the place in the original code where this behavior is encountered? Sorry if this should be obvious (in which case just tell me how I should have found it).

tdms/include/interpolation_methods.h
```diff
@@ -0,0 +1,42 @@
...
+                                 double a4, double a5, double a6, double a7);
+
+/**
+ * @brief Defines aliases for the flags that will determine the interpolation scheme to be used.
+ * 
+ * The choice of expansion is not arbitrary: for the band-limited cases, the expanded number can be used to tell us the range of Yee indicies we need to extract values from, to then interpolate.
+ * Similarly for cubic interpolation: the expanded number, offset by 7, fulfils the same role.
+ */
+enum interp_scheme
+{
```
t-young31, line 28:
```suggestion
enum interp_scheme {
```
we can probably clang-format everything but probably good to be consistent with the rest of the codebase (here and it lots of other places 😅)

tdms/src/interpolation_methods.cpp
```diff
@@ -0,0 +1,394 @@
...
+}
+
+/**
+ * @brief Given two interpolation schemes s0 and s1, performs evaluation of the expression "s0 > s1", in terms of optimality of the scheme.
+ *
+ * @param s0,s1 interp_schemes
+ * @return true If s0 is the optimal scheme.
+ * @return false If s1 is the optimal scheme. Also returned in the event of a tie.
+ */
+bool better_scheme(interp_scheme s0, interp_scheme s1) {
```
t-young31, line 357: This may be slightly more readable if the interp_scheme was a class, then you can have a method like `scheme0.is_better_than(scheme1)`

tdms/tests/unit/test_interpolate.cpp
t-young31: @willGraham01 did you find where this was occurs in iterator.cpp ?

tdms/CMakeLists.txt
```diff
@@ -78,6 +78,7 @@ target_sources(tdms_lib PUBLIC
         src/argument_parser.cpp
         src/fdtd_grid_initialiser.cpp
         src/interpolate.cpp
+        src/interpolation_methods.cpp
```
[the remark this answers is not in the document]
willGraham01: Plan is to eventually replace `interpolate.cpp` with `interpolate_Efield.cpp` and `interpolate_Hfield.cpp`. `interpolation_methods.cpp` will contain methods common to both, like implementing band-limited/cubic interpolation on a set of points, the `interp_scheme` enum (or class)

tdms/include/interpolation_methods.h
willGraham01: Agreed - VSCode defaults to that style so I'll be vigilant from now on :sweat_smile:

tdms/tests/unit/test_interpolate.cpp
willGraham01: An example of the original error check [is here](https://github.com/UCL/TDMS/blob/9a8a2291ec4783a320e1a9b75018d5e92bfb0082/tdms/src/interpolate.cpp#L144-L165) and use of the `k_l=0=k_u` case is [here](https://github.com/UCL/TDMS/blob/9a8a2291ec4783a320e1a9b75018d5e92bfb0082/tdms/src/iterator.cpp#L6071-L6073).
(The other TE/TM interpolation functions also have similar error checks, but I'll stick to talking about this specific instance as the others are similar.) 

It is not checked that `k_l < k_u`. However it looks like this would be taken care of implicitly, as the [`for` loops that follow](https://github.com/UCL/TDMS/blob/9a8a2291ec4783a320e1a9b75018d5e92bfb0082/tdms/src/interpolate.cpp#L168-L174) would be empty, so "nothing would be done". Whether this is the intended behaviour or not I wasn't sure, but I've preserved the behaviour by _not_ letting `checkInterpolationPoints` throw an error in the `k_l < k_u` case.

tdms/src/interpolation_methods.cpp
willGraham01: I debated it; went with an enum because besides this one function, I don't actually need any fancy functionality surrounding `interp_scheme`. Can convert it to a class if you'd prefer.

tdms/tests/unit/test_interpolate.cpp
t-young31: I would have thought `k_l <= k_u` is definitely the intended behaviour. probably worth adding that as an additional constraint(?)

tdms/src/interpolate.cpp
t-young31: this constraint can be implicit when we pass an `ElectricField` instance

tdms/src/interpolation_methods.cpp
t-young31: it might be worth it – as they can also then have their own interp_data attributes that are current defined at the top of each `interpolateTimeDomainE{x,y,z}` function (`{cubic,bandlimited}_interpolation` can also take 1 fewer argument)

tdms/src/interpolate_Efield.cpp
```diff
@@ -0,0 +1,136 @@
+#include "interpolate_Efield.h"
+#include "interpolation_methods.h"
+
```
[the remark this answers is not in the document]
willGraham01: There should only be max. 3 of them once we're done, and 2 of the planned ones (`Efield` and `Hfield`) might even be gobbled up by the field-classes anyway. We can always decide on this once we've got the bandlimited functions working and tested?

tdms/src/interpolation_methods.cpp
willGraham01: Will make the class and push the changes tomorrow morning :+1:

tdms/tests/unit/test_interpolate.cpp
willGraham01: Adding this check causes the tests to fail when running `run_pstd_bscan`; see https://github.com/UCL/TDMS/runs/7896933760?check_suite_focus=true.
At some point in the 2D simulation we need to allow j_l > j_u in order to not throw an error, because the interpolation functions rely on j_l = 2 > j_u = -2 being true.
For now, I've avoided adding this check until we're completely clear on what is relying on it. Also, the new `E-field` interpolation functions that act component-wise rather than mode-wise, won't depend on this behaviour.

tdms/src/interpolation_methods.cpp
willGraham01: Class now exists: https://github.com/UCL/TDMS/blob/5f470ae405af6f6cff8d520eaf8faf5d8a67466a/tdms/include/interpolation_methods.h#L92

tdms/tests/unit/test_interpolate.cpp
t-young31: exciting
very happy to leave out that check if that possibility will be removed later anyway 👍🏼

tdms/src/interpolation_methods.cpp
t-young31: having both `interpScheme` and `interp_scheme` with the latter generated with `scheme.get_scheme()` feels weird 😕 
I was thinking to have some const class instances with associated _values_ e.g

```c++
const auto BAND_LIMITED_0 = interpScheme(0);   // where 0 is the value
```
which then have a overloaded call `()` operator to evaluate whatever interpolation they do. Then a function like

```c++
interpScheme best_interpolation_scheme(int cells_in_direction, int cell_id){
...
}
```

apologies if this is foolish and not possible 😅

willGraham01: It's probably possible, but it boils down to where we want the "determine the scheme" part to come into it. It has been a while since I got this deep into C++ classes and overloads though :sweat_smile:.

The `interp_scheme` enum is there to make sure we can't request a scheme that doesn't exist, and just to make things a bit more readable than "use scheme 0". I guess we don't need it; the implicit converter/overloaded constructor method will still work if you pass it an `int`, and I can just add an error check if an invalid scheme is requested.

In its current state, creating an instance of the class immediately determines the interpolation scheme it will perform, which can then be run with the `.interpolate(double *v)` method. This means when we interpolate we do the following:
- Decide on a cell to interpolate
- Create an instance of the `interScheme` class - this also determines the best interpolation scheme and sets this class instance to use that scheme
- Gather the interpolation data we need
- Pass the data to `.interpolate()` (or an overloaded call operator).

With constant class instances all we're doing to do is move the "determine which scheme" part:
- Decide on a cell to interpolate
- Determine the appropriate scheme to use, and create a pointer (or similar) to the corresponding constant class instance.
- Gather the interpolation data
- Interpolate with `.interpolate()` or the overloaded call operator.

I guess the latter avoids us creating and destroying a class instance each time we interpolate, and we can remove the "determine the best scheme" process from the constructor. I just thought it'd be _safer_ to include this check in the constructor, as it means that the interpolation functions don't get cluttered with a bunch of `switch`es depending on which constant instance we need to point to?

t-young31: > The interp_scheme enum is there to make sure we can't request a scheme that doesn't exist, and just to make things a bit more readable than "use scheme 0".

Definitely agree that 'use scheme 0' is kinda useless! but I think having `best_interpolation_scheme` return a class instance ensures that it does exist though(?)

***

I like not creating a class for each call to `interpolate()`, also means we could get rid of some switch statements(?). Also it would be nice to remove the e..g `..(scheme - 6)...` expressions which aren't super readable.

***

Happy to approve/merge this is in and make a PR with a suggestion, if you think it's worth it

willGraham01: > I like not creating a class for each call to interpolate(), also means we could get rid of some switch statements(?). Also it would be nice to remove the e..g ..(scheme - 6)... expressions which aren't super readable.


No worries, let me fix this here as merging it in is going will likely create problems down the line (the syntax for testing will likely change quite a bit).

So proposed changes are:
`interpScheme` class constructor just takes a value (enum, int, whatever) corresponding to the scheme that is to be used. We create constant instances of these classes (and give them suitable names), one for each scheme.
`best_interpolation_scheme(cells_in_direction, cell_id)` returns (a pointer to?) the constant class instance corresponding to the best interpolation scheme to use. This function is called whenever we want to interpolate.

I'll also address the `scheme-6` stuff by making the variable name a bit more informative and/or writing a method for `interpScheme` to extract this info!

This changes the previous approach where `best_interpolation_scheme(cells_in_direction, cell_id)` is built-into the `interpScheme` constructor, and an instance of the class is created & destroyed each time we interpolate.

t-young31: sounds good!

> No worries, let me fix this here as merging it in is going will likely create problems down the line 

cool cool

> interpScheme class constructor just takes a value (enum, int, whatever) corresponding to the scheme that is to be used. 

I would be inclined for the constructor to only take its associated 'value' (how good it is number). Then have a const instance for each scheme currently in `interp_scheme`. Also should have its index as an attribute (or some similar name) so we could get rid of the `..(scheme - 6)...` like things and have `scheme.index` instead (where `scheme` is a interpScheme instance). They could then each have associated `interpolate()` functions and not have to have lots of switch statements that currently exist in e.g. `bandlimited_interpolation(...)`

> (a pointer to?)

Definitely [return a reference](https://www.tutorialspoint.com/cplusplus/returning_values_by_reference.htm)

willGraham01: > I would be inclined for the constructor to only take its associated 'value' (how good it is number).

Any strong reason for this preference? I personally prefer the other way - pass in the scheme identifier and then assign the value of the scheme, just because it's the scheme that provides the value, rather than the value dictating which scheme should be used?

t-young31: I thought it was the value which dictates what scheme will be used? to me that's the only thing that may change about an interpolation scheme. the index it acts on given some data and what it does are constant, the priority might change(?) e.g. if more interpolation schemes are added. I may be miss-understanding what you mean though 😕

willGraham01: I'm using "value" rather ambiguously to be fair - maybe "preference" is a better word. From the interpolation document, here's how I understand how we should choose the interpolation scheme:
- If we can use BLi with 4 data points either side, use this. [This scheme has the highest "value"]
- If we can use BLi with 3 data points one side, and 5 the other, use this
- BLi with 2 one side / 6 other
- BLi with 1 one side / 7 other
- Cubic to centre of 4 points
- Cubic with 1 data point on one side / 3 on the other side [Lowest "value"]
The difficulty is that EG BLi with 3 data points "to the left" and 5 "to the right" has the same value as BLi with 5 data points "to the left" and 3 "to the right". So currently; "value" does not uniquely determine which "scheme" to be applied, but "scheme" does determine the "value".

t-young31: >  maybe "preference" is a better word.

definitely, or `priority`. I guess that number in addition to whether it is possible to use that scheme determines if it will be used

willGraham01: Yeah, have rejigged the class now so hopefully things are a lot more readable (and compact). There's still an `enum scheme_value` to provide the `priority` of each scheme, but we now have constant class instances and `best_interpolation_scheme` returning references to them.
`interpolateTimeDomainE{x,y,z}` should be a lot more readable now too!

t-young31: love it. is `interpolateTimeDomainH{x,y,z}` getting the same treatment?

willGraham01: Yep but will do this on another branch + PR. Sadly won't be _as_ nice to read due to the whole 2D-interpolation thing, but should still be much more readable

t-young31 (approved):
squash and merge away 🚀

PR #64, opened 2022-08-09
Refactor iterator [1/N]

### LINKED ISSUE #63: Refactoring iterator.cpp
`mexFunction` in iterator.cpp is the best part of 7k lines. This should be broken out into smaller functions with structs/classes for the commonly grouped function arguments.

t-young31:
Works towards closing #63 . This PR:

- Adds a couple of simple classes for electric `E` and magnetic `H` fields
- Adds a `Timer` class suitable for printing intervals and recording the time between starts and ends. Maximises code reuse and more clearly expresses intention
- Adds a test of the `phasor_norm` functions (equiv. to `extractPhasorENorm` and `extractPhasorHNorm`)
- Adds namespace prefixes to header files removing the requirement for `using namespace std` before including headers
- Starts a `SimulationParameters` class to hold the full set of parameters required for the simulation
- Removes the unused `matlabio_globals.h` header
- Moves numeric.h to the correct directory
- Adds some comments for the simulation dimensionality
- Starts to replace 0 -> false and 1 -> true, for readability
- Implements some clang-tidy suggestions in numerical_derivative.cpp
- Uses the built in pi constant from `cmath`
- Deletes some dead code in `mesh_base.cpp`

_With a view to have a modest diff this will be continued in other PRs_ 
***
### Timings
On an M1 Pro Macbook. Note that there is a few % margin of error between reruns.

<details>
  <summary>Initial</summary>

```bash
$ export OMP_NUM_THREADS=1
$ time py.test ../tests/system/test_arc01.py 
real	1m57.158s
user	1m55.738s
sys	0m0.496s
```
</details>

<details>
  <summary>Final</summary>

```bash
real	1m57.463s
user	1m56.250s
sys	0m0.565s
```
</details>

***
### Dependencies

- #48


tdms/src/iterator.cpp
```diff
@@ -262,39 +244,20 @@ inline int min ( int a, int b ) { return a < b ? a : b; }
...
-
-
-void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
-{
+void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[]){
   
   /*Local variables*/
   #ifdef FDFLAG
-  int useCD = 1; //determines whether FDTD or PSTD is used
+  bool useCD = true;  // Use central difference derivatives
```
t-young31, line 252: could be a compile time flag

t-young31: applied on https://github.com/t-young31/ifdtd/tree/refactor_2

tdms/src/electric_field.cpp
```diff
@@ -0,0 +1,16 @@
...
+using namespace std;
+
+
+void ElectricField::add_to_angular_norm(double f, int n, int Nt, SimulationParameters &params) {
+  angular_norm += phasor_norm(f, n, params.omega_an, params.dt, Nt);
+}
+
+complex<double> ElectricField::phasor_norm(double f, int n, double omega, double dt, int Nt){
+  return f
+         * exp( fmod(omega*((double) (n+1))*dt, 2*dcpi) * I)
```
willGraham01, line 14: What's `I` set to (in the exponent)? It's not an input nor a class attribute?

tdms/src/numerical_derivative.cpp
```diff
@@ -43,7 +34,7 @@ void complex_mult_vec( fftw_complex *a, fftw_complex *b, fftw_complex *c, int le
 void init_diff_shift_op( double delta, fftw_complex *Dk, int N){
   //fprintf(stdout,"init_diff_shift_op, delta=%e\n",delta);
   //define an
-  double *an = (double *)malloc(sizeof(double) * N);
+  auto *an = (double *)malloc(sizeof(double) * N);
   if( (N % 2) == 0){//even case
     for(int i=0;i<=(N/2);i++)
       an[i] = (double) i;
```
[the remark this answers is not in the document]
t-young31: this was a clang-tidy suggestion in my IDE. apparently "avoids duplicating the type name". not a big deal but making my IDE happy makes me happy 🤣

tdms/src/magnetic_field.cpp
```diff
@@ -0,0 +1,16 @@
...
+using namespace std;
+
+
+void MagneticField::add_to_angular_norm(double f, int n, int Nt, SimulationParameters &params) {
+  angular_norm += phasor_norm(f, n, params.omega_an, params.dt, Nt);
+}
+
+complex<double> MagneticField::phasor_norm(double f, int n, double omega, double dt, int Nt){
+  return f
+         * exp( fmod(omega*((double) n + 0.5)*dt, 2*dcpi) * I)
```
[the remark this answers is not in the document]
t-young31: this might be a question for @prmunro

tdms/src/electric_field.cpp
t-young31: the imaginary unit defined in globals.h:

https://github.com/t-young31/ifdtd/blob/289fbdc50b7a441c5bce7c37d611bb6ba75c07c7/tdms/include/globals.h#L73

tdms/src/magnetic_field.cpp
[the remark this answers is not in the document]
prmunro: I don't think this is a problem with the time convention. The time convention means that to go from a frequency domain field Ex to a time domain field ex, the following equation applies: ex = real( Ex*exp(-i\omega t)), when going from time domain data to frequency domain data the opposite sign is used, since this is the inverse operation.

If it was a sign error,  the error metric would be much larger as this would return the complex conjugate of the required field quantity.

tdms/include/simulation_parameters.h
```diff
@@ -0,0 +1,10 @@
+
+class SimulationParameters{
```
t-young31, line 2: this class will be expanded significantly in later PRs

tdms/src/timer.cpp
```diff
@@ -0,0 +1,25 @@
...
+void Timer::start() {
+  start_time = omp_get_wtime();
+}
+
+void Timer::end() {
+  end_time = omp_get_wtime();
+}
+
+double Timer::delta_seconds() const {
+  return start_time - end_time;
```
t-young31, line 15:
```suggestion
  return end_time - start_time ;
```

tdms/tests/unit/test_fields.cpp
```diff
@@ -0,0 +1,57 @@
+#include "catch2/catch_test_macros.hpp"
+#include "complex"
+#include "field.h"
+
+using namespace std;
+
+
+inline bool is_close(double a, double b){
+  return abs(a - b) < 1E-10;
```
giordano, line 9: This definition is quite naive, it should at very least keep into account the scale of `a` and `b`.  `1e-100` is very much different from `1e-10` but with your definition they'd be equal.  Similarly, `1e100` and `1.0000000000000002e100` (the `double` number following `1e100`) are pretty close for any practical uses, but they'd be different according to this definition.  The docstring of Julia's [`isapprox`](https://docs.julialang.org/en/v1/base/math/#Base.isapprox) provides some details about how to better compare floating point numbers.

tdms/tests/unit/test_fields.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+inline bool is_close(complex<double> a, complex<double> b){
+  return is_close(a.real(), b.real()) && is_close(a.imag(), b.imag());
```
giordano, line 13: Also, Julia's `isapprox` always compares `abs(a - b)`, which would be slightly different than operating on real and imaginary parts separately as you're doing here, the difference is comparing a rectangle vs a circle, but this is probably less important than adjusting the scale.

tdms/include/globals.h
```diff
@@ -63,17 +55,13 @@ enum RCSType   { parallel , perpendicular };
 //			Mathematical Constants
 // **************************************
 
-
 const double dcpi  = 3.14159265358979;
```
giordano, line 58: It wasn't touched in this PR, but
```suggestion
const double dcpi  = 3.141592653589793;
```
is a more accurate double-precision representation of pi (any more digit would be redundant).  Alternatively, include `math.h` and use `M_PI` without reinventing the wheel.

tdms/include/field.h
```diff
@@ -0,0 +1,32 @@
+#include "complex"
```
giordano, line 1: Typically, system headers are included with `<...>`, headers of the current program with `"..."`: https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html#Include-Syntax.  In this case, `"complex"` would need to be `<complex>`.  The same everywhere else for all other headers.

tdms/src/iterator.cpp
```diff
@@ -1,49 +1,33 @@
...
- */
-
-/*---------------------------------------------------------------*/
-//                        INCLUDE section
-/*---------------------------------------------------------------*/
-#include <omp.h>
+#include "omp.h"
 #include "mat_io.h"
-#include <complex>
+#include "complex"
```
giordano, line 9: For example, this was right as `<complex>`.

tdms/src/iterator.cpp
```diff
@@ -4930,16 +4881,13 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
...
-      fprintf(stdout,"%.03e ",secs);
-      time_0=time_1;
+      timer.click();
     }
     /********************/
   
     //update terms for self consistency across scattered/total interface - E updates##
     if(sourcemode == sm_steadystate){//steadystate
-      commonPhase = exp(-I*fmod(omega_an[0]*time_H,2.*dcpi));
+      complex<double> commonPhase = exp(-I*fmod(omega_an[0]*time_H,2.*dcpi));
```
giordano, line 4890: I'm surprised you didn't use `auto` here :smile:

tdms/tests/unit/test_fields.cpp
t-young31: fair enough – will modify accordingly. used |a - b| because it's easier to understand at a glance than |a - b|/max(a, b), when it's not hit

tdms/tests/unit/test_fields.cpp
t-young31: not sure if comparing a circle is more or less intuitive to be honest, but will adjust. shouldn't change anything

tdms/include/globals.h
t-young31: sounds good 👍🏼

tdms/include/field.h
t-young31: thanks – did not know that. will modify

tdms/src/iterator.cpp
t-young31: auto _all_ the things 😉

tdms/include/globals.h
```diff
@@ -63,17 +56,13 @@ enum RCSType   { parallel , perpendicular };
...
 
 
 // **************************************
 //			Physical Constants
 // **************************************
 
-
 const double eo	     =   8.85400e-12;				// free space electric permitivity
 const double mo	     =   4.0 * dcpi * 1.0e-7;		// free space magnetic permeability
 const double light_v =   1.0 / sqrt(mo*eo);			// free spave light velocity
```
willGraham01, line 68:
```suggestion
const double light_v =   1.0 / sqrt(mo*eo);			// free space light velocity
```

tdms/include/globals.h
t-young31: or perhaps not: [build now fails on Windows](https://github.com/UCL/TDMS/runs/7968328592?check_suite_focus=true#step:10:40 )

giordano:
The reason why `M_PI` couldn't be found on Windows is because having to define the macro `_USE_MATH_DEFINES` before including `cmath` (or `math.h`) introduces an ordering problem: `globals.h` has to be included ***before*** `cmath` (or `math.h`) or any other header file which includes them (like `complex`), because when it gets to `globals.h` the `math.h` header file wouldn't be loaded again a second time, and the first time the definition of the math constants was skipped because `_USE_MATH_DEFINES` wasn't defined.

This patch should do the trick:
```diff
diff --git a/tdms/src/electric_field.cpp b/tdms/src/electric_field.cpp
index 0e02b64..f338606 100644
--- a/tdms/src/electric_field.cpp
+++ b/tdms/src/electric_field.cpp
@@ -1,5 +1,5 @@
-#include <complex>
 #include "globals.h"
+#include <complex>
 #include "field.h"
 
 using namespace std;
diff --git a/tdms/src/iterator.cpp b/tdms/src/iterator.cpp
index 01a6aff..f47e67e 100644
--- a/tdms/src/iterator.cpp
+++ b/tdms/src/iterator.cpp
@@ -4,6 +4,7 @@
  *                 such as phasor extraction etc. Works in both pulsed 
  *                 and steady state mode.
  ******************************************************************/
+#include "globals.h"
 #include <complex>
 #include <algorithm>
 #include <cstring>
@@ -15,7 +16,6 @@
 #include "numeric.h"
 #include "mesh_base.h"
 #include "numerical_derivative.h"
-#include "globals.h"
 #include "matlabio.h"
 #include "mesh_base.h"
 #include "timer.h"
diff --git a/tdms/src/magnetic_field.cpp b/tdms/src/magnetic_field.cpp
index 3c8a7e6..6f585d4 100644
--- a/tdms/src/magnetic_field.cpp
+++ b/tdms/src/magnetic_field.cpp
@@ -1,5 +1,5 @@
-#include <complex>
 #include "globals.h"
+#include <complex>
 #include "field.h"
 
 using namespace std;
diff --git a/tdms/src/numerical_derivative.cpp b/tdms/src/numerical_derivative.cpp
index 78e888c..85df4f2 100644
--- a/tdms/src/numerical_derivative.cpp
+++ b/tdms/src/numerical_derivative.cpp
@@ -1,7 +1,7 @@
+#include "globals.h"
 #include <cmath>
 #include <cstdlib>
 #include <fftw3.h>
-#include "globals.h"
 #include "numerical_derivative.h"
 #include "matlabio.h"
```

t-young31:
thanks for investigating! I'd be rather inclined to leave the it as is, because adding in a constraint on ordering header files is a little annoying, imo

giordano (approved):

prmunro (approved):
These changes all look reasonable

PR #74, opened 2022-08-19
[Time-Domain] Interpolation functions for the H-field

willGraham01:
Addresses #44, tasks:
- H-field interpolation schemes
- BLi interpolation unit tests

Adds functions that will perform the interpolation (across two dimensions) for the H-field in the time domain, using the framework developed for the E-field components.

We again employ three separate functions for each of the H-field components. There is even less benefit to a generic "interpolate component" function than in the E-field case, due to the need to interpolate in two separate dimensions when handling the H-field.

Also addresses a typo in `interpolateTimeDomainE{x,y,z}` - syntax around `scheme.index` should be `(scheme.index+1)` rather than `(scheme.index-1)`.

This pull request does not alter the functionality of the current `tdms` executable or codebase. This will be done in a future PR that simultaneously addresses #44 tasks:
- Add frequency-domain interpolation functions for E/H field components
- Unit tests for field-component interpolation
- Change interpolation calls in `exphasorssurface`, `exphasorsvolume`, and `fieldsample` to the new interpolation functions.


tdms/include/interpolate_Hfield.h
```diff
@@ -1,18 +1,48 @@
...
-Additionally, generalised functions for performing the 2D interpolation is not pretty code to look at (although is theoretically possible if we wanted a slight speedup).
-This forces us to pass in I,J,K to _every_ component function, rather than only the number of cells in the appropriate direction.
-
-If we want to interpolate all the H-field components, there are functions for this which call the individual component functions separately.
-*/
-
+/**
+ * @brief Interpolate the x-component of the magnetic field to the centre of Yee cell i,j,k
+ *
+ * Interpolation for the magnetic field must be done across two dimensions - first we must interpolate to obtain nearby field values parallel to the centre of the Yee cell, then we must interpolate again in another dimension to the centre of the Yee cell itself.
```
t-young31, line 4: maybe worth setting a line length limit in your IDE. although we can just clang-format everything in future

tdms/src/interpolate_Hfield.cpp
```diff
@@ -1,161 +1,179 @@
...
+            {
+                // gather the data for interpolating in the z dimension through cell i - (x_scheme.index + 1) + ii
+                other_dim_data[kk] = Hxy[k - (z_scheme.index + 1) + kk][j][i - (x_scheme.index + 1) + ii] + Hxz[k - (z_scheme.index + 1) + kk][j][i - (x_scheme.index + 1) + ii];
+            }
+            // interpolate in z to obtain a value in cell i - (x_scheme.index+1) + ii for the Hx field
+            // place this into the appropriate index in the data being passed to the x_scheme
+            pass_to_x_scheme[ii] = z_scheme.interpolate(other_dim_data);
+        }
     }
     else {
```
t-young31, line 38: there feels like there could be some more code reuse i.e. 

```c++
other_scheme = z_scheme.is_better_than(y_scheme) ? z_scheme : y_scheme
```

then the loop to set `pass_to_x_scheme`

t-young31 (commented):
Looks great thus far 🖌️

t-young31:
could do a rebase to remove the commit history in the old branch, but also 🤷🏼. will be squashed anyway

t-young31:
but it is worth fixing this clang warning:

```
/.../tdms/tests/unit/test_interpolate.cpp:118:11: warning: taking the absolute value of unsigned type 'bool' has no effect [-Wabsolute-value]
    CHECK(abs(v12 - interp2(v1, v2, v3, v4) <= tol));
```

looks like the `abs` bracket is in the wrong place. sorry for not spotting this in the review 🤦🏼

tdms/src/interpolate_Hfield.cpp
willGraham01: The difficulty is that the `other_dim_data` we need to extract also depends on the scheme we are employing; for example when we use `z_scheme` we need to extract in the `k` index [here](https://github.com/UCL/TDMS/blob/3a0d133c1ace849a3e9bcf926aace00417945ea8/tdms/src/interpolate_Hfield.cpp#L31) and when we use `y_scheme` we need to extract in the `j` index [here](https://github.com/UCL/TDMS/blob/3a0d133c1ace849a3e9bcf926aace00417945ea8/tdms/src/interpolate_Hfield.cpp#L46).

There's probably a clever way around this using pointers, which'll let us get some more code-reuse out of this. To get something going I figured I'd start with two separate loops for when `z` and `y` are better. Suggestions are welcome, pointers in C++ tend to mess with my head at the best of times :sweat_smile:

tdms/include/interpolate_Hfield.h
willGraham01: For some reason VSCode defaults this to 200 :sweat_smile: - have reduced it to a more managable 100 now

tdms/src/interpolate_Hfield.cpp
t-young31: ah I didn't spot that – may be more challenging that I envisaged! 

***

>  pointers in C++ tend to mess with my head at the best of times 😅
you and me both 😅

willGraham01:
> but it is worth fixing this clang warning:
> 
> ```
> /.../tdms/tests/unit/test_interpolate.cpp:118:11: warning: taking the absolute value of unsigned type 'bool' has no effect [-Wabsolute-value]
>     CHECK(abs(v12 - interp2(v1, v2, v3, v4) <= tol));
> ```
> 
> looks like the `abs` bracket is in the wrong place. sorry for not spotting this in the review 🤦🏼

Have fixed this one - tests are still passing and now I'm even checking the right condition :sweat_smile: .

giordano (approved):
I didn't find much to say about this.  Nice work

tdms/src/interpolate_Hfield.cpp
```diff
@@ -1,161 +1,235 @@
...
-        interp_scheme z_scheme = determineInterpScheme(K, k);
-        if (better_scheme(z_scheme, CUBIC_INTERP_MIDDLE))
-        {
-            // using a BLi scheme in z, assign memory we need
-            other_dim_cell_data = (double *)malloc(8 * sizeof(double));
+}
+
+void interpolateTimeDomainHz(double ***Hzx, double ***Hzy, int i, int j, int k, int I, int J, double *Hz)
+{
+    /* There is a choice in the order of interpolation; either interpolate in x first, then y, or vice-versa.
```
t-young31, line 155: this is a lot of comments to maintain. is there a way to write something generic suitable for all `interpolateTimeDomainH{x, y, z}` ?

tdms/include/interpolate_Hfield.h
```diff
@@ -1,18 +1,54 @@
...
+ * Interpolation for the magnetic field must be done across two dimensions.
+ * The stored Hy-component at position (i+Dx, j, k+Dz) is associated to the Yee cell (i,j,k).
+ * To interpolate Hy to the centre of cell (i,j,k), we must therefore interpolate in the x-direction, then the z-direction, or vice-versa.
+ *
+ * @param[in] Hyx,Hyz Split components of the Yee cell
+ * @param[in] i,j,k Yee cell index
+ * @param[in] I,K Number of Yee cells in the i,k directions respectively
+ * @param[out] Hy Interpolated value of the Hy field at centre of Yee cell i,j,k
+ */
+void interpolateTimeDomainHy(double ***Hxy, double ***Hxz, int i, int j, int k, int I, int K, double *Hx);
```
t-young31, line 27:
```suggestion
void interpolateTimeDomainHy(double ***Hxy, double ***Hxz, int i, int j, int k, int I, int K, double *Hy);
```

tdms/include/interpolate_Hfield.h
```diff
@@ -1,18 +1,54 @@
...
+ * Interpolation for the magnetic field must be done across two dimensions.
+ * The stored Hz-component at position (i+Dx, j+Dy, k) is associated to the Yee cell (i,j,k).
+ * To interpolate Hz to the centre of cell (i,j,k), we must therefore interpolate in the x-direction, then the y-direction, or vice-versa.
+ *
+ * @param[in] Hzx,Hzy Split components of the Yee cell
+ * @param[in] i,j,k Yee cell index
+ * @param[in] I,J Number of Yee cells in the i,j directions respectively
+ * @param[out] Hz Interpolated value of the Hz field at centre of Yee cell i,j,k
+ */
+void interpolateTimeDomainHz(double ***Hxy, double ***Hxz, int i, int j, int k, int I, int J, double *Hx);
```
t-young31, line 41:
```suggestion
void interpolateTimeDomainHz(double ***Hxy, double ***Hxz, int i, int j, int k, int I, int J, double *Hz);
```

tdms/include/interpolate_Hfield.h
```diff
@@ -1,18 +1,54 @@
...
 
-void interpolateTimeDomainHx(double ***Hxy, double ***Hxz, int i, int j, int k, int I, int J, int K, double *Hx);
-void interpolateTimeDomainHy(double ***Hxy, double ***Hxz, int i, int j, int k, int I, int J, int K, double *Hx);
-void interpolateTimeDomainHz(double ***Hxy, double ***Hxz, int i, int j, int k, int I, int J, int K, double *Hx);
+/**
+ * @brief Interpolate the H-field to the centre of Yee cell i,j,k
+ *
+ * @param[in] Hxy,Hxz,Hyx,Hyz,Hzx,Hzy Split components of the Yee cell
+ * @param[in] i,j.k Index of the Yee cell to interpolate to the centre of
+ * @param[in] I,J,K Number of Yee cells in the i,j,k directions (respectively)
```
t-young31, line 48: @dstansby suggested e.g. `nI, nK, nK` were better variables names and I'm inclined to agree `nX` implies a number of x(?)

tdms/src/interpolate_Hfield.cpp
```diff
@@ -1,161 +1,235 @@
...
+    // determine the y-direction scheme
+    const interpScheme &y_scheme = best_interp_scheme(J, j);
 
-    // this interpolation scheme will determine how many times we need to interpolate in the other dimension, and the range of x-Yee cells to consider
-    int i_low, i_high, d_low, d_high;
-    // we will also need to store the result of the interpolation in the first dimension
-    double *first_interp_vals;
-    double *other_dim_cell_data;
+    // this data will be passed to the second interpolation scheme
+    double pass_to_second_scheme[8];
```
t-young31, line 26: maybe a slightly more consistent variable name might be good?  i.e. `data_for_second_scheme` rather than `pass_to_second_scheme`

tdms/tests/unit/test_interp_fns.cpp
```diff
@@ -0,0 +1,292 @@
+# include "catch2/catch_test_macros.hpp"
+# include "interpolation_methods.h"
+
+# include <algorithm>
+# include <cmath>
+# include <iostream>
+
+using namespace std;
+
```
t-young31, line 9: loving the tests so far – were you going to add some that test the `interpolateTimeDomain` functions?

willGraham01: I have another (local) branch for these tests. I have a couple of tests working already, interpolating some known functions and checking the error in the output matches the error obtained from having MATLAB's `interp` (which BLi is based of). So I'm confident that BLi _is_ doing what it should!

Mostly just waiting on today's (5th) meeting for some ideas surrounding a "BLi beats Cubic interpolation" test, to confirm that BLi is providing an _improvement_ in the parameter regimes we expect.

tdms/src/interpolate_Hfield.cpp
willGraham01: I'll break out the whole thing at the top of the file in a generic setting, then use shorter, more contextual inline comments in the functions themselves

tdms/include/interpolate_Hfield.h
willGraham01: Will do - I'll also change the E-field functions and `interpolation_methods` too whilst I'm at it

samcunliffe:
Sorry for the conflucts 😞

willGraham01:
All good now - I just can't merge correctly :sweat_smile:

tdms/include/interpolation_methods.h
```diff
@@ -105,7 +103,11 @@ class interpScheme {
 
         /* END FETCH METHODS */
         
-        // stores the index-offset that we require when extracting data from the field component arrays
+        /* Stores the index-offset that we require when extracting data from the field component arrays
+        Let F be the (split field) Yee cell components of some field component, and suppose we are interested in interpolating this field to the centre of the cell with index u.
+        Then the data point F[u - (index+1)] is plays the role of v[0] in the interpolation scheme.
```
prmunro, line 108: Is v[0] the same as v0 in the above ASCII diagrams, which explain interp1, interp2, etc. In particular:

 * v0    v1    v2    v3
 * o     o  x   o    o

tdms/include/interpolation_methods.h
```diff
@@ -48,13 +48,11 @@ double interp3(double *v);
...
- * @param K Number of elements in the k direction of the FDTD grid
+ * @param nI,nJ,nK Number of elements in the i,j,k (respectively) direction of the FDTD grid
  *
  * @throws runtime_error In the event that the limits of field extraction are outside the FDTD grid
  */
-void checkInterpolationPoints(int i_l, int i_u, int j_l, int j_u, int k_l, int k_u, int I, int J, int K);
+void checkInterpolationPoints(int i_l, int i_u, int j_l, int j_u, int k_l, int k_u, int nI, int nJ, int nK);
 
 /**
  * @brief Defines our order of preference for the use of the various schemes.
```
prmunro, line 58: Would it be possible to have some additional comments explaining what is meant by "interp position = 0", etc., below. I think I understand what is meant, but it would be good to make this explicit. Upon further reading it is clear, however, I think a small additional comment somewhere would make this clearer.

tdms/src/interpolation_methods.cpp
```diff
@@ -30,29 +30,29 @@ double interp3(double *v)
...
     }
     else if (k_l < 2)
     {
         throw runtime_error("Interpolation error: k_l too small");
     }
-    else if (k_u > K - 2)
+    else if (k_u > nK - 2)
     {
         throw runtime_error("Interpolation error: k_u too large");
     }
```
prmunro, line 58: Below, int the case of CUBIC_INTERP_FIRST, and the other cubic schemes, why is:

last_nonzero_coeff = 4;

when it looks (to my current understanding) as though it should be:

last_nonzero_coeff = 3;

tdms/src/interpolate_Hfield.cpp
```diff
@@ -1,161 +1,231 @@
...
+            // interpolate in z to obtain a value for the Hy field at position (cell_i+Dx, j, k)
+            // place this into the appropriate index in the data being passed to the x_scheme
+            data_for_second_scheme[ii] = z_scheme.interpolate(data_for_first_scheme);
         }
-        else
+        // now interpolate in the x-direction to the centre of Yee cell (i,j,k)
+        *Hy = x_scheme.interpolate(data_for_second_scheme);
+    }
+    else
+    {
```
samcunliffe, line 139: Marginal and/or nit-picky comment, but [Google style](https://google.github.io/styleguide/cppguide.html#Conditionals) prefers...
```c++
if (condition) {                   // no spaces inside parentheses, space before brace
  DoOneThing();                    // two space indent
  DoAnotherThing();
} else if (int a = f(); a != 3) {  // closing brace on new line, else on same line
  DoAThirdThing(a);
} else {
  DoNothing();
}
```
what editor do you use? Can you try `clang-format` (or similar). 

In case of vim there's a [Google C++ style file](https://github.com/vim-scripts/google.vim) you can use. In case of VSCode it should happen by default(?). In case of emacs, you're on your own 😛

prmunro (changes requested):
Hi Tom and Sam, I'm reasonably clear on this code, but I have a few queries and requests for additional comments that would help me at this stage. If the requested comments are present and I've missed them, of course just point me to them. Looking good though.

tdms/src/interpolate_Hfield.cpp
samcunliffe: Obviously apply style rules iff appropriate for clarity etc.

tdms/tests/unit/test_interp_determination.cpp
```diff
@@ -81,12 +91,15 @@ TEST_CASE("determineInterpScheme: correct interpolation chosen") {
...
-    for(int i=4; i<=N-4; i++) {CHECK(best_interp_scheme(N, i).get_priority() == BAND_LIMITED_3);}
-    CHECK(best_interp_scheme(N, N-3).get_priority() == BAND_LIMITED_4);
-    CHECK(best_interp_scheme(N, N-2).get_priority() == BAND_LIMITED_5);
-    CHECK(best_interp_scheme(N, N-1).get_priority() == BAND_LIMITED_6);
-
-    /* If 4<=N<=8 we can still fall back on cubic interpolation 
+    for (int i = 4; i <= N - 4; i++)
+    {
+        CHECK(best_interp_scheme(N, i).get_priority() == BAND_LIMITED_3);
+    }
```
samcunliffe, line 97: More trivial suggestions:
```suggestion
    for (int i = 4; i <= N - 4; i++)
        CHECK(best_interp_scheme(N, i).get_priority() == BAND_LIMITED_3);
```
Or maybe one line? (Whatever your preferred editor/linter suggests.)

tdms/tests/unit/test_interp_fns.cpp
```diff
@@ -0,0 +1,292 @@
...
+    CHECK(abs(v23 - interp1(v1, v2, v3, v4)) <= tol);
+    CHECK(abs(v34 - interp3(v1, v2, v3, v4)) <= tol);
+}
+
+/**
+ * @brief The hard-coded numerical values for the interpolation constant should all sum to the same value
+ *
+ * Note - the coefficients are not required to sum to unity!
+ */
+TEST_CASE("bandlimited_interpolation: coefficient sum")
```
samcunliffe, line 80: A kind-of filippant suggestion...
```suggestion
TEST_CASE("bandlimited_interpolation: check that the interpolation constant values all sum to the same value")
```
Probably don't accept mine, but expand on `coefficient sum`. A few dittos for your test cases ff.

tdms/tests/unit/test_interp_fns.cpp
```diff
@@ -0,0 +1,292 @@
...
+    CHECK(abs(v12 - interp2(v1, v2, v3, v4)) <= tol);
+    CHECK(abs(v23 - interp1(v1, v2, v3, v4)) <= tol);
+    CHECK(abs(v34 - interp3(v1, v2, v3, v4)) <= tol);
+}
+
+/**
+ * @brief The hard-coded numerical values for the interpolation constant should all sum to the same value
+ *
+ * Note - the coefficients are not required to sum to unity!
+ */
```
samcunliffe, line 79: Do these comments get rendered in doxygen? If not, do we want them? (maybe not).

samcunliffe (approved):
No functionality-based comments from me. I leave Peter to pass an eye over the method/physics.

Should we (or does it make sense to) add a system test to cover this (I mean a full blown call to tdms input → output as expected).

samcunliffe:
@dstansby it doesn't hurt to have you go over this too if you have time.

tdms/tests/unit/test_interp_fns.cpp
```diff
@@ -0,0 +1,292 @@
...
+# include <cmath>
+# include <iostream>
+
+using namespace std;
+
+/**
+ * @brief In the case when cubic interpolation is to be used, check that all polynomial fields up to cubic order are interpolated exactly (to within machine error)
+ *
+ */
+TEST_CASE("interp: cubic interpolation is exact")
```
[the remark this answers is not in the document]
willGraham01: `Catch2` even allows some kind of "tagging" of tests too if we want to group them up nicely across files.

Happy to defer to more informative strings regardless though :+1:

tdms/src/interpolate_Hfield.cpp
willGraham01: This is the result of VS Code's hideous default formatting that I have now overwritten :sweat_smile:

The style I've been _writing_ is:
```c++
if (condition) {       // no spaces inside parentheses, space before brace 
    DoOneThing();      // 4 space indent 
    DoAnotherThing(); 
} 
else if (int a = f(); a != 3) {  // closing brace on new line, else on same line 
    DoAThirdThing(a); 
} 
else {
    DoNothing(); 
}
```
however what you're seeing here is the result of that being run through VSCode's formatting after editing a seemingly unrelated line of code...

tdms/tests/unit/test_interp_fns.cpp
willGraham01: I don't know what doxygen does regarding these comments - I thought I'd include them for readability reasons nonetheless.

willGraham01: If they make it into the doxygen that's a bonus, but if they don't chances are the person running the tests will be a developer with access to the source code anyway. It's also less of a problem since we're making the `TEST_CASE` strings more verbose too.

samcunliffe: They get ignored. But could be included with adding [tdms/tests/units to the Doxyfile](https://github.com/UCL/TDMS/blob/main/doc/Doxyfile#L22), and adding a 
```c++
/**
 * @file test_interp_fns.cpp
 * Test the blah blah
 */
```
At the top.

... My question was really: In your opinion, is it worth to document the tests in Doxygen? I'm of two minds:

* [No] Because we should stick to the working code in the docs.
* [Yes] Because developers are going to need to understand the tests.

Probably out of scope for this PR.

tdms/src/interpolate_Hfield.cpp
```diff
@@ -1,161 +1,231 @@
...
+    // determine the y-direction scheme
+    const interpScheme &y_scheme = best_interp_scheme(nJ, j);
+
+    // this data will be passed to the second interpolation scheme
+    double data_for_second_scheme[8];
+    // this data will hold values for the interpolation in the first interpolation scheme
+    double data_for_first_scheme[8];
+
+    // which of z_scheme and y_scheme is WORSE? We will interpolate in this direction SECOND
+    if (z_scheme.is_better_than(y_scheme))
```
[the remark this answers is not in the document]
willGraham01: The reason for the separation is because we have to read the data from different axes of the `E{x,y,z}` etc arrays. [There was this discussion](https://github.com/UCL/TDMS/pull/74#discussion_r951168214) about the same thing - there probably is a way to reduce the code repetition, but I can't see an immediate and clear way.

tdms/include/interpolate_Hfield.h
```diff
@@ -1,28 +1,54 @@
...
+ * Interpolation for the magnetic field must be done across two dimensions.
+ * The stored Hx-component at position (i, j+Dy, k+Dz) is associated to the Yee cell (i,j,k).
+ * To interpolate Hx to the centre of cell (i,j,k), we must therefore interpolate in the y-direction, then the z-direction, or vice-versa.
+ *
+ * @param[in] Hxy,Hxz Split components of the Yee cell
+ * @param[in] i,j,k Yee cell index
+ * @param[in] nJ,nK Number of Yee cells in the j,k directions respectively
+ * @param[out] Hx Interpolated value of the Hx field at centre of Yee cell i,j,k
  */
-#pragma once
```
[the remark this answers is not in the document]
willGraham01: It did indeed get mangled. Have reinstated `#pragma once`. The big docstring chunk _was_ meant to go, being replaced by the function-docstrings and a brief file docstring at the top.

tdms/include/interpolation_methods.h
willGraham01: No problem - I've added a comment at the end of this comment block, and have included an ASCII diagram similar to those for the cubic interpolation functions.

tdms/include/interpolation_methods.h
willGraham01: Yes - don't know why I included the square bracers around `[0]` here!

tdms/src/interpolation_methods.cpp
willGraham01: They should be 3, this is me being guilty of copy/paste without checking everything

tdms/src/interpolate_Hfield.cpp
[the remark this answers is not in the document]
samcunliffe: Sorry for the duplicated comment!
But good to know I'm thinking along similar lines to Tom.

tdms/src/interpolate_Efield.cpp
```diff
@@ -2,56 +2,56 @@
...
-    const interpScheme &scheme = best_interp_scheme(J, j);
+    const interpScheme &scheme = best_interp_scheme(nJ, j);
     // prepare input data - if using a cubic scheme we have reserved more memory than necessary but nevermind
     double interp_data[8];
 
     // now fill the interpolation data
-    // j + scheme.index-1 is the index of the Yee cell that plays the role of v0 in the interpolation
-    // note that scheme.index-1 appears here because Yee cells associate field values "to the right" of their centre (IE, the field component is at a position with a higher coordinate value than the Yee cell centre)
+    // j - (scheme.index+1) is the index of the Yee cell that plays the role of v0 in the interpolation
+    // note that scheme.index+1 apears here because Yee cells associate field values "to the right" of their centre (IE, the field component is at a position with a higher coordinate value than the Yee cell centre)
```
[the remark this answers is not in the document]
willGraham01: The `index` field [is documented with a note](https://github.com/UCL/TDMS/blob/80c1b1e8a74952718bb17472a4397c7f9a5b3db0/tdms/include/interpolation_methods.h#L112-L117) concerning why this `+1` later appears. `index` is definitely not the most informative name choice, but equally I couldn't think of anything succinct to name this variable. Using your suggestion I'll change its name to `number_of_datapoints_to_left` or something, which better describes it's new role. `BAND_LIMITED_CELL_ZERO` will have a rather value for this variable, but otherwise it'll probably make things clearer.

My original idea was that `index` matches the `interpolation position` labeling that is set up and described [here](https://github.com/UCL/TDMS/blob/80c1b1e8a74952718bb17472a4397c7f9a5b3db0/tdms/include/interpolation_methods.h#L64-L68). Then, the `+1` seems fairly natural to me since the Yee cells always associate field components "to the right" of their centres (IE, the centre of a Yee cell is never has a coordinate greater than those of the position of any field components associated to that cell). A similar logic also holds for the cubic schemes. It also means we can avoid adjustments to the code when we want to interpolate "to the left" of cell 0 (`BAND_LIMITED_CELL_ZERO`).

tdms/src/interpolate_Efield.cpp
```diff
@@ -2,56 +2,56 @@
...
-    const interpScheme &scheme = best_interp_scheme(I, i);
+    const interpScheme &scheme = best_interp_scheme(nI, i);
     // prepare input data - if using a cubic scheme we have reserved more memory than necessary but nevermind
     double interp_data[8];
 
     // now fill the interpolation data
-    // i + scheme.index-1 is the index of the Yee cell that plays the role of v0 in the interpolation
-    // note that scheme.index-1 appears here because Yee cells associate field values "to the right" of their centre (IE, the field component is at a position with a higher coordinate value than the Yee cell centre)
+    // i - (scheme.index+1) is the index of the Yee cell that plays the role of v0 in the interpolation
+    // note that scheme.index+1 apears here because Yee cells associate field values "to the right" of their centre (IE, the field component is at a position with a higher coordinate value than the Yee cell centre)
```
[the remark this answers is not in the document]
willGraham01: I believe @prmunro already updated the PSTD_Interpolation_Scheme document after our last in-person meeting. The Overleaf document has the correct notation (albeit indexing starts from 1 and not from 0). 

With the change from `index` to `number_of_datapoints_to_left`, the mysterious `+1`s disappear so we can avoid this confusion too.

prmunro (approved):
Thanks for making those changes, using number_of_..._left is much more understandable to me.

PR #75, opened 2022-08-22
Use CMake `CTest` module to choose whether to build tests

giordano:
This is a more standard way to enable/disable CMake tests, instead of (ab)using the build type.


t-young31 (approved):
🚀

Issue #70, opened 2022-08-10
Remove MATLAB dependency


t-young31:
The MATLAB dependency from TDMS can be removed, which will (a) simplify the build, (b) allow running on machines without a MATLAB licence and (b) allow static linking such that no runtime paths need to be set and a compiled binary distributed.

***
### Decisions

- Tensor library vs. nested std::vector vs. raw arrays

***
### Dependencies

- #63

samcunliffe:
> will (a) simplify the build, (b) allow running on machines without a MATLAB licence and (c) allow static linking such that no runtime paths need to be set and a compiled binary distributed.

+ (d) _should_ be architecture-independent. (ARM64 builds on Mac OS?)

PR #80, opened 2022-09-05
E- and H-field interpolation tests, frequency domain interpolation

willGraham01:
# Interpolation Tests for the $E$- and $H$-fields.

Dependent on #74.

Completes the following task in #44:
- Tests for BLi interpolation methods.

Adds a BLi vs cubic placeholder test, which at present **is expected to fail**, since the setup and function to be interpolated do not satisfy the conditions under which we expect BLi to be "better" than cubic. To be updated with a suitable function and/or domain setup.

Addresses the following task in #44:
- Unit tests for `interpolate{TimeDomain,FrequencyDomain}{E,H}{x,y,z}` functions.

Adds tests against MATLAB's `interp` function, which BLi is based off. Using predefined functional forms for the components of the two fields, and a particular grid setup to mimic the Yee cells and stored field components. The maximum pointwise-error of the interpolation to the Yee cell centres is computed, and compared to that from MATLAB.
The script to compute the corresponding errors in MATLAB can be included in the repository if so desired - at present the output value is manually placed into the acceptance tolerance variable.

## Also to be included

Frequency-domain interpolation functions.
These will simply take in complex data and pass the `.real` and `.imag` components to the real-valued interpolation methods.

Completes the following task in #44:
- Write interpolation methods for the E- and H- fields in the frequency domain. Compared to the time-domain, this scheme must be able to handle complex data.


samcunliffe (commented):
Right. A few comments for clarity. I admit to being a little bit lost in the actual interpolation stuff. That is not necessary a bad thing. 

Worth asking Peter into this review? (presumably better to merge in the namespaces PR and clear up the diff, first).

.github/workflows/macos_tests.yml
```diff
@@ -58,26 +54,24 @@ jobs:
...
-
       - name: Run TDMS system tests
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
         # run system tests in mac only on pushes to main (they are a bit slow)
         working-directory: ${{runner.workspace}}/build
         shell: bash
         run: |
           export OMP_NUM_THREADS=1
-          py.test ${GITHUB_WORKSPACE}/tdms/tests/system/ -s -x
+          py.test ${GITHUB_WORKSPACE}/tdms/tests/system/ -s -x
```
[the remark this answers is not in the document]
willGraham01: Could well be my VSCode's obsession with _removing_ newlines too.

tdms/include/interpolate_Efield.h
```diff
@@ -46,3 +90,16 @@ void interpolateTimeDomainEField(double ***Exy, double ***Exz, double ***Eyx,
                                  double ***Eyz, double ***Ezx, double ***Ezy, 
                                  int i, int j, int k, int nI, int nJ, int nK, 
                                  double *Ex, double *Ey, double *Ez);
+/**
+ * @brief Interpolate the E-field to the centre of Yee cell i,j,k
+ *
+ * Frequency domain corresponds to interpolating complex-valued fields.
```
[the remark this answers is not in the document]
willGraham01: These just handle the interpolation - solution of the system happens somewhere else in `iterator.cpp`. The actual method of interpolation is pretty much identical regardless of whether or not we are using complex or real data (freq/time domain), the only difference being whether or not we have split-field information or not. Hence my templating everything here :sweat_smile:

tdms/tests/unit/test_field_interpolation.cpp
```diff
@@ -0,0 +1,451 @@
...
+                                             cell_centre[1],
+                                             cell_centre[2] + 0.5 * cellDims[2]);
+                Hyz[kk][jj][ii] = 0.;
+                Hzx[kk][jj][ii] = Hcomponent(cell_centre[0] + 0.5 * cellDims[0],
+                                             cell_centre[1] + 0.5 * cellDims[1],
+                                             cell_centre[2]);
+                Hzy[kk][jj][ii] = 0.;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
willGraham01: I've opened #125 for this, have put the relevant information I have in there too so I don't forget my original though process.

tdms/include/interpolation_methods.h
```diff
@@ -1,11 +1,14 @@
 /**
  * @file interpolation_methods.h
- * @brief Cubic interpolation functions.
+ * @author William Graham (ccaegra@ucl.ac.uk)
+ * @brief InterpScheme class methods and supporting functions
  * 
- * Needed to  preserve interpolate.cpp current functionality prior to testing!
+ * Non InterpScheme methods are required to preserve functionality whilst testing new schemes
```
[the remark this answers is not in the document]
willGraham01: Have put a reminder in #126. Shouldn't be difficult

tdms/include/interpolation_methods.h
```diff
@@ -128,14 +131,18 @@ class interpScheme {
...
-         * @param v Sample datapoints to use in interpolation
+         * For slight speedup, the actual sum performed loops over those i such that
+         * 0 <= first_nonzero_coeff <= i <= last_nonzero_coeff <= 7.
+         * 
+         * @param v Sample datapoints to use in interpolation; v[0] should be the first of 8 values
          * @param offset [Default 0] Read buffer from v[offset] rather than v[0]
          * @return double Interpolated value
          */
         double interpolate(const double *v, const int offset = 0) const;
+        std::complex<double> interpolate(const std::complex<double> *v, const int offset = 0) const;
```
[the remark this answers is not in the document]
willGraham01: VSCode seems to understand it, haven't actually checked doxygen though :/ Either way, these are now function templates because they do _literally the same thing_, so only one docstring is needed.

tdms/src/interpolate_Efield.cpp
```diff
@@ -1,66 +1,113 @@
...
     // prepare input data - if using a cubic scheme we have reserved more memory than necessary but nevermind
     double interp_data[8];
 
     // now fill the interpolation data
     // j - scheme.number_of_datapoints_to_left is the index of the Yee cell that plays the role of v0 in the interpolation
-    for (int ind = scheme.first_nonzero_coeff; ind <= scheme.last_nonzero_coeff; ind++)
-    {
-        interp_data[ind] = Eyx[k][j - scheme.number_of_datapoints_to_left + ind][i] + Eyz[k][j - scheme.number_of_datapoints_to_left + ind][j];
+    for (int ind = scheme.first_nonzero_coeff; ind <= scheme.last_nonzero_coeff; ind++) {
+        interp_data[ind] = Eyx[k][j - scheme.number_of_datapoints_to_left + ind][i] + Eyz[k][j - scheme.number_of_datapoints_to_left + ind][i];
```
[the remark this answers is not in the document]
willGraham01: This is easiest to explain in pictures - grab me in BH/HH next week if what's below isn't clear. Some setup below:

General information about the algorithm is in the `PSTD_interpolation_scheme` document. The caviat is that the field values associated to the Yee cells lie on the faces of the cells, rather than at their centres (hence the need for interpolation). So the field value stored in `Eyx[k][j][i]` is the field value _associated_ to the cell `(i,j,k)`, _but_ isn't the value of the field at the Yee cell centre. In fact, if `(x_i, y_j, z_k)` are the coordinates of the centre of cell `(i,j,k)`, then `Eyx[k][j][i]` is the split-field value at position `(x_i, y_j + delta_j, z_k)` where `delta_j` is half the length in the $y$-direction of the Yee cell. There are similar rules in Peter's document for the other field components.

As such, we interpolate "between the Yee cell _values_" (the `Eyx[k][j][i]` etc) to obtain the field at the Yee cell centres. Since the `Eyx[k][j][i]` are all evenly spaced, we have a situation like below:
```
X     X     X     X     X     X     X     X
                     o
```
Here, the `X`s are the positions at which we have the `Eyx[k][j][i]`, whilst `o` is the centre of one particular Yee cell that we want to interpolate into. Since the `Eyx[k][j][i]` are evenly-spaced, and `o` is at the midpoint of two `X`s (being the centre of a particular Yee cell), we can interpolate using the Yee cell values "at the edges" (the `Eyx[k][j][i]`) to obtain the fields at the centre of the Yee cells.

Now the answer to your question :sweat_smile: 
We're interpolating to the centre of cell `(i,j,k)` in the `y`-direction, so we don't change the `k` and `i` indices in `Eyx` when we extract the datapoints we need. Let's suppose we have a situation like the above, so we then need to look at 4 datapoints either side of the `o`. But remember that the datapoint immediately to the _right_ of `o` is the value `Eyx[k][j][i]`. So, the 8 datapoints we need must come from the Yee cells whose y-indices are:
```
j-4, j-3, j-2, j-1,   j , j+1, j+2, j+3
 X    X    X    X     X    X    X    X
                   o
```
Hence, this weird `for` loop. `interp_data` is an array of length 8 that'll store the datapoints that we use to interpolate with. `ind` is an index running from 0 to 7 that we can use to extract the 8 consecutive values from `Eyx` we need. `scheme.number_of_datapoints_to_left` in this instance would be 4 - we need 4 datapoints (`j-4` to `j-1`) before `Eyx[k][j][i]` to perform this scheme.

The `InterpScheme` class sets the private value of `number_of_datapoints_to_left` when an instance is created, because in general we might be in a situation where we don't have 4 cells either side of `o` to use in our interpolation. For example,
```
X     X     X     X     X     X     X     X
         o             
```
in the above, `number_of_datapoints_to_left` will be 2, and we will take datapoints from cells `j-2` through to `j+5`. Conversely, a situation like this
```
X     X     X     X     X     X     X     X
                                       o
```
will have `number_of_datapoints_to_left` being 7.

Finally, you also highlighted that `ind` doesn't go from 0 to 7, but rather from `scheme.first_nonzero_coeff` to `scheme.last_nonzero_coeff`. Again, these are just fail-safes for when we are in these non-symmetric situations. When we are near the edges, some interpolation schemes require fewer than 8 datapoints (since the some of the scheme's coefficients are 0), so this prevents seg-faults that might happen if we _always_ tried to get 4 datapoints either side of `o`. It also means that when we fall back on cubic interpolation, we don't always try and read double as many values as we need!

tdms/tests/unit/matlab_benchmark_scripts/benchmark_test_BLi_vs_cubic_validation.m
```diff
@@ -0,0 +1,92 @@
...
+The function we use is a Cauchy distribution; f(x) = 1/(10x^2+1), over the
+range x\in[-2,2]
+
+Neither cubic nor BLi should be the superior interpolation method for all
+datapoint-separation  distances.
+
+Results are printed out for each cell size to stdout.
+%}
+clear; 
+close all;
```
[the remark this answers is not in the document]
willGraham01: MATLAB script starting boilerplate. MATLAB sometimes saves workspaces on shutdown, so this accounts for that

tdms/tests/unit/test_field_interpolation.cpp
```diff
@@ -0,0 +1,451 @@
...
+ * @brief Test the interpolation of the E-field components to the centre of the Yee cells
+ *
+ * Each component of the E-field will take the form
+ * E{x,y,z}(xx,yy,zz) = sin(2\pi zz) * exp(-yy^2) * ( 1./ (10xx^2+1) ).
+ *
+ * We test both the Fro- and slice-norm metrics, since interpolation only happens along one axis
+ */
+TEST_CASE("E-field interpolation check") {
+    cout << fixed << scientific << setprecision(8);
+    cout << "===== Testing E-field BLi =====" << endl;
```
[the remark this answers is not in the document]
willGraham01: It does, but apparently there's no way [to set it once and then keep writing](https://github.com/gabime/spdlog/issues/1019). Will change to spdlog though - forgot to after my own debugging.

tdms/tests/unit/test_interpolation_functions.cpp
```diff
@@ -1,89 +1,136 @@
...
+    // array that will be passed to Interp_scheme::interpolate
+    double interp_data[4];
 
     // constant field
     double c0 = 3.1415;
+    interp_data[0] = c0;
+    interp_data[1] = c0;
+    interp_data[2] = c0;
+    interp_data[3] = c0;
+
```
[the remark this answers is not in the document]
willGraham01: Good point. Going to leave `c0` in though, just in case we decide to test using a different (more physically relevent?) polynomial.

prmunro:
Just some comments following quoted text for clarity

> # Interpolation Tests for the E- and H-fields.
> Dependent on #74 | Dependent on #108 (namespacing issues)
> 
> Completes the following tasks in #44:
> 
> * [x]  Closes [Unit tests for BLi interpolation methods. #111](https://github.com/UCL/TDMS/issues/111)
> * [x]  Closes [Unit tests for `interpolate{TimeDomain,FrequencyDomain}{E,H}{x,y,z}` functions. #110](https://github.com/UCL/TDMS/issues/110)
> 
> Tests the performance of the BLi interpolation methods against the MATLAB function the scheme is based off. MATLAB scripts which output the error for particular field choices are included in the `tests/unit/matlab_benchmark_scripts` directory. The output error values are currently hard-coded into the respective tests; this is not an ideal solution, however in light of #70 is an adequate testing method until the MATLAB dependency question is answered.
> 
> The tests themselves are against MATLAB's `interp` function. Using predefined functional forms for the components of the two fields; and a particular grid setup to mimic the Yee cells and stored field components, the maximum pointwise-error of the interpolation to the Yee cell centres is computed, and compared to that from MATLAB.
> 
> Tests of the BLi method for complex data has been added to `test_interpolation_functions`. A field-wide time-domain interpolation test has been added to `test_field_interpolation.cpp`.
> 
> Completes the following task in #44:
> 
> * [x]  Closes [Write interpolation methods for the E- and H- fields in the frequency domain. Compared to the time-domain, this scheme must be able to handle complex data. #112](https://github.com/UCL/TDMS/issues/112)
> 
> The codebase currently uses `mxArray`s to encode complex fields, and the functions `mxInterpolateFieldCentral{E,H}` to handle the interpolation. The new functions use `complex` as the datatype for the inputs. This is to be consistent with `test_numerical_derivative` and `field.h`, and to use a complex datatype we are importing already (and which won't be removed by #70). However, these functions will be are unused by the code, as the `mxInterpolate` functions are still used until `iterator`'s refactor is complete.
> 
> It is also worth noting that interpolation in the frequency domain _does not_ appear to use split-field components, unlike that in the time domain. As such, the functionality of the time domain and frequency domain functions are slightly different (which doesn't permit templating).
> 
> A test on the frequency-domain interpolation has been added to `test_field_interpolation.cpp`.
> 
> ## Key Files to Check:
> * `test_field_interpolation.cpp`
> * `test_BLi_vs_cubic_interpolation.cpp`
> * `interpolate_{E,H}.h` (although these changes just turn the field-interpolation functions into template functions, so we can use `double` (time domain) or `std::complex<double>` (freq domain) data in each.

"The output error values are currently hard-coded into the respective tests; this is not an ideal solution": agreed it is not ideal, however, calling Matlab functions within TDMS is a definite no. I'm content with the hard-coding for now, since it could be amended relatively easily at a later date if a suitable function was found, either implemented by us or someone else.

" It is also worth noting that interpolation in the frequency domain _does not_ appear to use split-field components, unlike that in the time domain": this is correct since frequency domain data is stored in non-split format. Split components are summed together to form the non-split field prior to executing a partial sum contributing to the temporal discrete Fourier transform.

All comments above make good sense are sensible and make good sense (for what it's worth).

tdms/include/interpolate_Hfield.h
```diff
@@ -1,11 +1,49 @@
...
+Hz  | (0.5, 0.5, 0.0) .* (Dx, Dy, Dz)
+
+where Dx, Dy, Dz are the dimensions of the Yee cell.
+This requires us to interpolate twice to recover (any of the) field components at the centre of Yee cell i,j,k.
+
+Henceforth, let {a,b,c} = {x,y,z} be some 1-to-1 assignment of the co-ordinate axes to the labels a,b,c.
+The values Da, Db, Dc are the corresponding permutation of Dx, Dy, Dz.
+Suppose that we wish to interpolate the Ha field component to the centre a particular Yee cell.
+
+We will use the notation (a_i,b_j,c_k) for the Yee cell indices, although bear in mind that this does not reflect the order the indices appear in the code.
```
prmunro, line 25: @willGraham01 , could you add just a little more explanation here? I don't quite follow the notation. Is the logic below that you will always access Ha, and therefore, the field component of interest will be chosen by setting a (i.e., if interested in Hz, we assign a=z)? In this case it makes sense.

tdms/tests/unit/matlab_benchmark_scripts/benchmark_test_field_interpolation_E.m
```diff
@@ -0,0 +1,164 @@
...
+     BENCHMARK FOR E-FIELD INTERPOLATION METHODS
+
+Tests the interpolation accuracy for the E-field components, when
+interpolated in each direction to the centre of a Yee cell - benchmarking
+test_field_interpolation::TEST_CASE("E-field interpolation check").
+
+The E-field used will have identical field components, each of which are
+given by the expression:
+
+E{x,y,z}(xx,yy,zz) = sin(2\pi zz) * exp(-yy^2) * ( 1./ (10xx^2+1) ),
```
prmunro, line 11: It would be more realistic if the function was something more like:

E{x,y,z}(xx,yy,zz) = sin(2\pi xx) * exp(-xx^2) *sin(2\pi yy) * exp(-yy^2)* sin(2\pi zz) * exp(-zz^2)

tdms/tests/unit/matlab_benchmark_scripts/benchmark_test_field_interpolation_H.m
```diff
@@ -0,0 +1,193 @@
...
+
+Tests the interpolation accuracy for the H-field components, when
+interpolated in each direction to the centre of a Yee cell - benchmarking
+test_field_interpolation::TEST_CASE("H-field interpolation check").
+
+The H-field used will have identical field components, each of which are
+given by the expression:
+
+H{x,y,z}(xx,yy,zz) = cos(0.5\pi zz) * exp(-yy^2) * ( 1./ (5xx^2+1) ),
+
```
prmunro, line 12: See comment above about alternative function. Also, has 0.5\pi been chosen intentionally? This is sub-optimal as (if I understand correctly), we have at least 16 points per sinusoidal oscillation, which is too many to warrant band-limited interpolation

tdms/include/interpolate_Hfield.h
willGraham01: Yep this was the intention - we always try and extract `Ha`. That's also why there are [comments like this](https://github.com/UCL/TDMS/blob/fbc348174285a3874dfdc5f9d5ee356c8a5511a2/tdms/include/interpolate_Hfield.h#L59) after each function is defined. If we are interested in `Hz`, as you say we use `a=z`. Then it doesn't really whether we choose `b=x, c=y` or `b=y, c=x` since we interpolate in each of these directions anyway. The reason for the abstract notation is just to highlight that whichever direction we choose for `a`, the two dimensions we interpolate in are the _other_ dimensions `b` and `c`.

Originally I had this comment at the start of each function, explicitly using Hx, Hy, and Hz and moving the roles around as necessary. That resulted in a huge file-bloat and a lot of repeated information, so we abstracted it and moved it outside the functions into one comment at the beginning of the file.

prmunro (changes requested):
This broadly looks great. I have made some comments which I think need some attention. Apologies if I have misunderstood something!

tdms/tests/unit/matlab_benchmark_scripts/benchmark_test_field_interpolation_E.m
willGraham01: Thanks - I'll update the tests to use the more realistic suggestions.
Just to confirm, that's the field
$$E_x(x,y,z) = \sin(2\pi x)e^{-x^2}, \quad E_y(x,y,z) = \sin(2\pi y)e^{-y^2}, \quad E_z(x,y,z) = \sin(2\pi z)e^{-z^2},$$
IE each component is sine multiplied by a Gaussian? (Rather than each component being the product of the three expressions above).

.github/workflows/macos_tests.yml
```diff
@@ -22,7 +22,6 @@ jobs:
       matrix:
         os:
           - macos-12
-          # TODO: can change to macos-latest (but it's currently aliased to macos-11)
```
samcunliffe, line 25: Sorry, one more thing: could you leave this please?

samcunliffe (approved):
Approving, for my part.
Obv. subject to Peter's change requests as above.

tdms/tests/unit/matlab_benchmark_scripts/benchmark_test_field_interpolation_H.m
willGraham01: $0.5\pi$ is me incorrectly translating my paper notes -> code. I've also taken a similar form for the $H$-field as for the $E$-field you suggested above. The tests have been updated with the new functional forms.

tdms/tests/unit/matlab_benchmark_scripts/benchmark_test_field_interpolation_E.m
prmunro: Actually, it's more realistic of Ex varied as a function of y and/or z, and similarly for the other components. Just because of the transverse nature of electromagnetic waves.

tdms/include/interpolate_Hfield.h
```diff
@@ -1,11 +1,49 @@
...
+For example; if a = y, b = x, c = z, then the value of Ha at cell (a_i,b_j,c_k) is accessed via Ha[c_k][a_i][b_j], due to the interchanging of the x and y directions.
+Similarly; we will write Ha[a_i, b_j, c_k] to refer to the value of Ha associated to cell (a_i,b_j,c_k).
+
+Suppose now that we have selected cell a_i,b_j,c_k to interpolate to the centre of.
+We must interpolate in the b-direction, then c-direction, or vice-versa.
+For optimal accuracy we must determine the best interpolation scheme we can use in each direction at cell (a_i,b_j,c_k), and use the WORSE scheme second.
+
```
prmunro, line 32: Can I just check how you determine which is the worst scheme?

tdms/tests/unit/matlab_benchmark_scripts/benchmark_test_field_interpolation_E.m
willGraham01: Something cyclic like
$$E_x(x,y,z) = \sin(2\pi y)e^{-y^2}, \quad E_y(x,y,z) = \sin(2\pi z)e^{-z^2}, \quad E_z(x,y,z) = \sin(2\pi x)e^{-x^2},$$
for example?

tdms/include/interpolate_Hfield.h
prmunro: Thanks, makes sense

tdms/tests/unit/matlab_benchmark_scripts/benchmark_test_field_interpolation_E.m
prmunro: Yes, that's better

prmunro (approved):

PR #82, opened 2022-09-05
Add a Doxyfile, and first gh action to publish.

samcunliffe:
Right, this feels a little bit like elastic bands and sticky tape, but here goes...

This PR introduces a new github action, which runs doxygen over tdms/include and tdms/src, then commits the contents of doxygen's output folder to the `gh-pages` branch. The next step (at time of writing is turned off in the settings for this repo) is to automatically publish the `gh-pages` branch... actually this is done by another action triggered by a new commit to `gh-pages`.

So our action to create the branch, followed by a default github action to publish the pages.

* [Action in the marketplace](https://github.com/marketplace/actions/doxygen-github-pages-deploy-action)
* [PoC here on my private fork](https://samcunliffe.me/TDMS/)
    - For example [this page](https://samcunliffe.me/TDMS/hierarchy.html), happens to be some of the way through Tom's refactoring, and therefore has some interesting class structure.
* [This action which committed to to the `gh-pages` branch](https://github.com/samcunliffe/TDMS/actions/workflows/doxygen-gh-pages.yml).

There may well be a smarter way to do this. Has anyone done this before/better?

Ideas for improvement 
----------------------
1. We might actually want the README.md to be in the root of our gh page. So:
    ```
    ➤ tree -L 1
    .
    ├── README.md
    └── doxygen-output
    ```
    Which means users get a nice welcomming page first. And we add a suitably clear link from our README to the developer docs.

2. [Nice class inheritance diagrams are possible](https://www.doxygen.nl/manual/examples/diagrams/html/inherits.html)


.github/workflows/doxygen-gh-pages.yml
```diff
@@ -0,0 +1,18 @@
+name: Doxygen GitHub Pages Deploy Action
+
+on:
+
+  # push:
```
t-young31, line 5: triggering on a main push seems like what we want?

t-young31 (commented):
Looks great to me. I've not used doxygen really at all. but I like the PoC so 👍🏼

t-young31 (approved):

giordano (approved):
Looks good to me!

I'm not super familiar with how doxygen works, so I have a couple of questions:
1. Is the Doxygen file a kind of configuration file? I was surprised by the many lines, but then I saw it's mostly comments
2. Is there a possibility that changes to the code can break building the docs? In that case we may want to run this also on pull requests

samcunliffe:
>     1. Is the Doxygen file a kind of configuration file? I was surprised by the many lines, but then I saw it's mostly comments

Yes, and yes it's huge. We could delete all of the comments (but I can never remember all of the options). The important line is [911](https://github.com/UCL/TDMS/pull/82/files#diff-a663b96f43877059a341a68f25be3489f38e1c6401fc5b23c2ee55182b32e924R911).

>     2. Is there a possibility that changes to the code can break building the docs? In that case we may want to run this also on pull requests

It's possible to delete or change the docstrings to break the documentation. Doxygen might throw warnings but stull make the pages.

t-young31:
> We could delete all of the comments (but I can never remember all of the options)

That would probably be my preference, not that it really matters

doc/Doxyfile
```diff
@@ -0,0 +1,26 @@
...
+PROJECT_BRIEF          = "Time Domain Maxwell Solver"
+# Blank == no logo
+PROJECT_LOGO           =
+# Blank == Write output to current dir.
+OUTPUT_DIRECTORY       =
+INPUT                  = tdms/include tdms/src
+FILE_PATTERNS          = *.cpp \
+                         *.h \
+                         *.py
+GENERATE_LATEX         = NO
```
giordano, line 26: Missing newline 
```suggestion
GENERATE_LATEX         = NO

```
Sorry, I'll always bang people for this

samcunliffe:
> Your site is live at http://github-pages.ucl.ac.uk/TDMS/

(I forgot to check the https option... will be https on next deployment 🤞 ).

PR #85, opened 2022-09-07
[CI] Add macos testing

samcunliffe:
Got it working on 🍎 !

Thanks for @giordano 's top-tip to use [this tmate action](https://github.com/marketplace/actions/debugging-with-tmate) to help debugging.

Unfortunately it's yet another separate build with a YML file that is 90% the same as the Linux and Windows ones. But on the plus: we get another badge.

Good news: it found a genuine (minor) bug. 🐛

### DIFF
--- a/.github/workflows/macos_tests.yml
+++ b/.github/workflows/macos_tests.yml
@@ -1,0 +1,82 @@
+name: MacOS tests
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+  release:
+  workflow_dispatch:
+
+jobs:
+  tests:
+    name: Test - ${{ matrix.os }} - Build type ${{ matrix.build_type }} - Build testing ${{ matrix.build_testing }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-11
+          - macos-12
+          # TODO: can change to macos-latest (but it's currently aliased to macos-11)
+        build_type:
+          - Release
+        build_testing:
+          - ON
+          - OFF
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+
+      - name: Setup cache
+        id: cache-test-data
+        uses: actions/cache@v3
+        with:
+          path: ${{ github.workspace }}/tdms/tests/system/data
+          key: test_data
+
+      - name: Set up MATLAB
+        uses: matlab-actions/setup-matlab@v1
+
+      - name: Install dependencies
+        run: |
+          brew install fftw llvm libomp python3
+
+      - name: Install Python dependencies
+        shell: bash
+        run: |
+          pip3 install -r ${GITHUB_WORKSPACE}/tdms/tests/requirements.txt
+
+      - name: Create Build Environment
+        run: |
+          cmake -E make_directory ${{runner.workspace}}/build
+
+      - name: Configure TDMS
+        shell: bash
+        working-directory: ${{runner.workspace}}/build
+        env:
+          CXXFLAGS: -Werror
+        run: |
+          cmake ${GITHUB_WORKSPACE}/tdms -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON\
+           -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}\
+           -DBUILD_TESTING=${{ matrix.build_testing }}\
+           -DOMP_ROOT=/usr/local/opt/llvm/ -DCXX_ROOT=/usr/local/opt/llvm
+
+      - name: Build TDMS
+        working-directory: ${{runner.workspace}}/build
+        shell: bash
+        run: |
+          cmake --build . --config ${{ matrix.build_type }}
+
+      - name: Run TDMS unit tests
+        if: matrix.build_testing == 'ON'
+        working-directory: ${{runner.workspace}}/build
+        run: |
+          export OMP_NUM_THREADS=1
+          ctest -C ${{ matrix.build_type }} --extra-verbos
+
+      - name: Run TDMS system tests
+        working-directory: ${{runner.workspace}}/build
+        shell: bash
+        run: |
+          export OMP_NUM_THREADS=1
+          py.test ${GITHUB_WORKSPACE}/tdms/tests/system/ -s -x
--- a/.github/workflows/macos_tests.yml
+++ b/.github/workflows/macos_tests.yml
@@ -15,7 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-11
           - macos-12
           # TODO: can change to macos-latest (but it's currently aliased to macos-11)
         build_type:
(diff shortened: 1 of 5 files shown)


t-young31 (approved):
🍎 🚀 

My only comment is: perhaps ~3h of CI time is a lot per run for this. I wonder if, like the Windows CI, we should just check a successful compile on `pull_request` (and run everything on `push: main`)

samcunliffe:
> 🍎 🚀
> 
> My only comment is: perhaps ~3h of CI time is a lot per run for this. I wonder if, like the Windows CI, we should just check a successful compile on `pull_request` (and run everything on `push: main`)

Oh agreed. I didn't notice that the system tests were at ~35mins each. (Why so slow?)

.github/workflows/macos_tests.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Run TDMS unit tests
+        if: matrix.build_testing == 'ON'
+        working-directory: ${{runner.workspace}}/build
+        run: |
+          export OMP_NUM_THREADS=1
+          ctest -C ${{ matrix.build_type }} --extra-verbos
+
+      - name: Run TDMS system tests
+        working-directory: ${{runner.workspace}}/build
```
samcunliffe, line 78:
```suggestion
      - name: Run TDMS system tests
        if: ${{ github.ref == 'refs/heads/main' }}
        # run system tests in mac only on pushes to master (they are a bit slow)
        working-directory: ${{runner.workspace}}/build
```

samcunliffe: OK this works. Is the strategy with you, @giordano ?

giordano: I agree the much longer wall time isn't great, but do we have an intuition of why tests on macOS are much slower than the rest?

Also, to reduce the usage of resources, can we perhaps test only on `macos-latest`?  Although not impossible, we'd need to be particularly unlucky to hit some version-specific peculiarities for this kind of software.

.github/workflows/macos_tests.yml
```diff
@@ -0,0 +1,82 @@
...
+        with:
+          path: ${{ github.workspace }}/tdms/tests/system/data
+          key: test_data
+
+      - name: Set up MATLAB
+        uses: matlab-actions/setup-matlab@v1
+
+      - name: Install dependencies
+        run: |
+          brew install fftw llvm libomp python3
```
giordano, line 42: Is `llvm` strictly necessary?  I thought that installing libomp would suffice, but I haven't tried myself.

.github/workflows/macos_tests.yml
```diff
@@ -0,0 +1,82 @@
...
+        working-directory: ${{runner.workspace}}/build
+        shell: bash
+        run: |
+          cmake --build . --config ${{ matrix.build_type }}
+
+      - name: Run TDMS unit tests
+        if: matrix.build_testing == 'ON'
+        working-directory: ${{runner.workspace}}/build
+        run: |
+          export OMP_NUM_THREADS=1
```
giordano, line 74: Instead of setting `OMP_NUM_THREADS` manually in each step (and I'd use
```yaml
        env:
          OMP_NUM_THREADS: 1
```
instead), I think you can also set the variable globally for the entire job.  This has also the advantage of being shell-independent (not that on macOS specifically we'd use much different shells, but I'm thinking in the optic of unifying the build scripts).

.github/workflows/macos_tests.yml
t-young31: I've had a small amount of experience with running MacOS jobs on gh runners and it seems like the hardware is just much slower

> Also, to reduce the usage of resources, can we perhaps test only on macos-latest? 

+1 on this

.github/workflows/macos_tests.yml
t-young31: tangential to this point: it may be nice to have a test e.g. just `test_arc_01.py `that runs with 2 threads, to test the parallelisation hasn't been broken?

giordano: Sounds good.  It could be done with an additional matrix entry, and then you can do
```yaml
env:
  OMP_NUM_THREADS: ${{ matrix.num_threads }}
```
We do it for example at https://github.com/Team-RADDISH/ParticleDA.jl/blob/410605a29b17d2ec2f3e86848e62a4ab6ff73102/.github/workflows/ci.yml#L12-L13

.github/workflows/macos_tests.yml
samcunliffe: `macos-latest` is `macos-11`. I would actually rather test 12. Hence my leaving the TODO.
Let me remove 11.

.github/workflows/macos_tests.yml
samcunliffe: > Instead of setting `OMP_NUM_THREADS` manually in each step (and I'd use
> 
> ```yaml
>         env:
>           OMP_NUM_THREADS: 1
> ```
> 
> instead), I think you can also set the variable globally for the entire job. This has also the advantage of being shell-independent (not that on macOS specifically we'd use much different shells, but I'm thinking in the optic of unifying the build scripts).

Ah. This was just brainlessly copied over from the Wndows one which was the starting point.

samcunliffe: > Sounds good. It could be done with an additional matrix entry, and then you can do
> 
> ```yaml
> env:
>   OMP_NUM_THREADS: ${{ matrix.num_threads }}
> ```
> 
> We do it for example at https://github.com/Team-RADDISH/ParticleDA.jl/blob/410605a29b17d2ec2f3e86848e62a4ab6ff73102/.github/workflows/ci.yml#L12-L13

I would do this in a separate issue/PR (to avoid snowballing). And let's do it to the linux tests (only)?

.github/workflows/macos_tests.yml
```diff
@@ -0,0 +1,82 @@
...
+
+jobs:
+  tests:
+    name: Test - ${{ matrix.os }} - Build type ${{ matrix.build_type }} - Build testing ${{ matrix.build_testing }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-11
```
samcunliffe, line 18:
```suggestion
```

.github/workflows/macos_tests.yml
samcunliffe: Huh. [You're right](https://github.com/samcunliffe/TDMS/runs/8300142473?check_suite_focus=true).

samcunliffe: Oh. I needed to do it on my local M2 machine because: architectures.

.github/workflows/macos_tests.yml
samcunliffe: > but do we have an intuition of why tests on macOS are much slower than the rest?

Seems to be an actual issue: actions/runner-images#3885

Issue #30, opened 2022-07-08
Add more system tests


t-young31:
Currently there are only PSTD tests, at least one FDTD test should also be included, and others.

- [ ] FDTD
- [x] steadystate = true
- [x] exdetIntegral = true
- [x] exi is present

samcunliffe:
Kinda high priority if it's blocking the refactor.

prmunro:
Here is a test of steady state behaviour. The script arc_08/run_pstd_bscan.m generates two output files:

arc_08/out/pstd_sph_steady.mat
arc_08/out/pstd_fs_steady.mat

which can be considered the correct result. Note that issue https://github.com/UCL/TDMS/issues/93#issue-1374674142 needs to be resolved before the TDMS version of the code will produce output that agrees with these files. 

run_pstd_bscan.m requires a particular directory structure as evident from the addpath statements. 

The test can be downloaded from UCL dropbox (too large for direct upload):

    You will need the claim passcode '5d0a7097' 
to retrieve the file(s).  

    It is up to you to scan the downloaded file(s) for viruses. Most anti-virus applications will do this automatically, but you should check.

    The file(s) will be available for 10 days.


*** Details for retrieval: ***
    
pick-up URL: https://wwwapps-live.ucl.ac.uk/cgi-bin/dropbox/dropbox-https.cgi?state=pickup_info&id=3d2c36cb
claim id: 3d2c36cb
file basename: arc_08.zip
file length: 563336149
file content_type: application/zip
description:

samcunliffe:
Thanks! I have it. 

Files arc01-03.zip were uploaded to zenodo and the [system tests grab the data then run tdms over them](https://github.com/UCL/TDMS/blob/main/tdms/tests/system/test_arc01.py#L9).

samcunliffe:
Wall times:
```
py.test ../test_arc08.py  6251.60s user 305.00s system 722% cpu 15:07.20 total
py.test ../test_arc03.py  1241.86s user 55.81s system 564% cpu 3:49.90 total
```

Running tdms over arc08 (in/pstd_fs_steady.mat and in/pstd_sph_fs_steady.mat) is a lot slower than, for example, [arc03](https://github.com/UCL/TDMS/blob/main/tdms/tests/system/test_arc03.py). Probably this is expected(?)  Does it make sense to down-sample here? I mean, just for speeding up the automatic test. Or can we live with this?

prmunro:
[...]

PR #88, opened 2022-09-09
Refactor [2/N]

t-young31:
Continues #63 

I'd strongly recommend we add some more comprehensive system tests before merging this.

***
### Guidelines for reviewers
These commits have been reviewed as separate PRs on my fork previously. If this PR warrants further review I'd strongly recommend reviewing the individual commits, with the exception of the one that just clang-formats the entire `iterator.cpp` file.

***
### Extensions

- Convert the cerr 'prints' to spdlog logging calls
- Cary on the refactor past the start of the main loop
- Change the non-matlab arrays (defined in `arrays.h`) to something more robust
- Use the defined `.clang-format` file to reformat the entire code base (ideally as a precommit hook)
- Add _many_ more unit tests (this refactor largely focused on extracting classes and structs that have matlab-dependent constructors, thus didn't warrant testing, if they're going to be rewritten)

***
### Dependencies

- #64 
- #30


.clang-format
```diff
@@ -0,0 +1,66 @@
+# Generated from CLion C/C++ Code Style settings
```
samcunliffe, line 1: Did you mean to commit this?
We will have a unified clang-format settings, so it's not a bad idea IMO.

samcunliffe (approved):
Just checking/reminding myself of what's going on here (and double checking for anything I might've missed). 

→ I've already reviewed all of the changes in little chunks: so approving here too.

.clang-format
t-young31: yep, but feel free to remove the comment/modify accordingly. strong in favour of clang-formatting the whole code base

samcunliffe: Ah, I meant the whole file. A common clang-format config is a good idea then.

prmunro (approved):

tdms/src/arrays.cpp
```diff
@@ -0,0 +1,401 @@
...
+  }
+
+  matrix = cast_matlab_2D_array(mxGetPr(ptr), n_rows, n_cols);
+  this->n_cols = n_cols;
+  this->n_rows = n_rows;
+}
+
+Pupil::~Pupil() {
+  free_cast_matlab_2D_array(matrix);
+  matrix = nullptr;
```
t-young31, line 194: this is an appalling hack – please can we get rid of matlab soon

samcunliffe (approved):

PR #90, opened 2022-09-12
[Doc] Add doxygen-style comments to the easy/obvious places.

samcunliffe:
This is probably the most exciting pull-request you've ever reviewed.

Adding some easy doxygen comment strings to the files that aren't likely to conflict with Tom's refactor, or at least not in any non-trivial way. Very low-priority, but please do check my understanding e.g. in numerical_derivative.h. Picky language comments are very very welcome.

I also ran [codespell](https://github.com/codespell-project/codespell) over everything 😎 .


tdms/include/mesh_base.h
```diff
@@ -1,3 +1,8 @@
+/**
+ * @file mesh_base.h
+ * @brief TODO: what are these functions
```
willGraham01, line 3: I think these are the functions that are meant to allow for extraction of the field over a given surface. The user can specify (EG a plane) to extract the field on, and these are the functions which are meant to mesh it.
Don't quote me on that though!

giordano (approved):
I wasn't asked to review this, but it seemed easy enough :slightly_smiling_face:   I haven't read all the comments in detail, but the PR looks overall good to me.

I only have one comment: since `#pragma once` is non-standard, is it worth mentioning somewhere (README?  dev docs?) that building the program requires a compiler supporting it?  However according to [Wikipedia](https://en.wikipedia.org/wiki/Pragma_once#Portability) most compilers should actually accept this directive (I verified the Fujitsu compiler supports it too).

tdms/src/argument_parser.cpp
```diff
@@ -5,8 +5,6 @@
 using namespace std;
 
 
-ArgumentParser::ArgumentParser() = default;
```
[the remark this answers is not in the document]
samcunliffe: Yes, exactly.
And I was always taught to put them in the header so if needed: I claim it should go back in there.

samcunliffe:
> I only have one comment: since `#pragma once` is non-standard, is it worth mentioning somewhere (README? dev docs?) that building the program requires a compiler supporting it? However according to [Wikipedia](https://en.wikipedia.org/wiki/Pragma_once#Portability) most compilers should actually accept this directive (I verified the Fujitsu compiler supports it too).

Yeah: [exactly](https://github.com/UCL/TDMS/blob/main/doc/developers.md?plain=1#L28).

giordano:
Cool, I clearly missed that :grin:   Perfect!

tdms/include/numerical_derivative.h
```diff
@@ -1,5 +1,50 @@
...
-void init_diff_shift_op( double delta, fftw_complex *Dk, int N);
-void first_derivative( fftw_complex *in_pb_pf, fftw_complex *out_pb_pf,
-		       fftw_complex *Dk, int N, fftw_plan pf, fftw_plan pb);
+
+/**
+ * @brief Multiply two arrays of complex numbers element-wise.
+ *
+ * Performs the complex multiplication of every element of a with every element
+ * of b. Stores the result in c. All arrays must be of equal length.  Used
+ * internally by first_derivative.
```
willGraham01, line 15:
```suggestion
 * Performs element-wise complex multiplication of the array a with the array b.
 * a and b must be must be of equal length.
 * Stores the result in c.
 * Internally used by first_derivative.
```
Just felt the "complex multiplication of every element of a with every element of b" implied some kind of outer product was happening.

tdms/include/numerical_derivative.h
```diff
@@ -1,5 +1,50 @@
...
+                      int len);
+
+/**
+ * @brief Initialise the coefficients required to simultaneously perform
+ * differentiation and shifting by amount delta, using a forward and backward
+ * FFT.
+ *
+ * @param delta The fraction of the spatial step.
+ * @param Dk The coefficients.
+ * @param N The number of elements in Dk.
```
willGraham01, line 32:
```suggestion
 * @param[in] delta The fraction of the spatial step.
 * @param[out] Dk Buffer to write the coefficients to.
 * @param[in] N The number of elements in Dk.
```

tdms/include/numerical_derivative.h
```diff
@@ -1,5 +1,50 @@
...
+ * @note in_pb_pf must be the buffer which is the input for both plans pf and pb.
+ * Likewise, out_pb_pf must be the output for both plans pf and pb.
+ *
+ * @param[in] in_pb_pf The buffer containing the data to be differentiated.
+ * 	    @warning This buffer will be overwritten as part of the computation.
+ * @param[out] out_pb_pf The buffer which will contain the computed derivative.
+ * @param Dk The coefficients.
+ * @param N Number of elements in buffers.
+ * @param pf The plan for forward FFT.
+ * @param pb The plan for backward FFT.
```
willGraham01, line 47:
```suggestion
 * @param[inout] Dk Buffer to write the coefficients for performing differentiation and shifting in Fourier space.
 * @param[in] N Number of elements in buffers.
 * @param[in] pf The plan for forward FFT.
 * @param[in] pb The plan for backward FFT.
```
I don't think that the plans `pb` and `pf` are overwritten at any point. `Dk` is also `inout` by technicality - we're passing in a buffer we don't mind overwriting, and I'm not sure if we ever use the coefficients again (although this might need a check).

tdms/include/openandorder.h
```diff
@@ -1,12 +1,51 @@
...
 
+/**
+ * @brief Iterate through the matrix names and assign the pointer to each matrix
+ * into the appropriate entry of pointers
+ * 
+ * @param expected 
+ * @param actual 
+ * @param pointers 
+ */
 void assign_matrix_pointers(MatrixCollection &expected, MatFileMatrixCollection &actual, const mxArray **pointers);
```
willGraham01, line 51: I'm not familiar with what this function should be doing either - does it just compare names of arrays that were read in to some reference list?

tdms/src/mesh_base.cpp
```diff
@@ -765,7 +765,7 @@ void crossProduct(int v1[], int v2[], int *v1crossv2){
 }
 
 
-//mex -v mesh_base.cpp ../matlabio3/matlabio.cpp -DMIOFLAG=matlab  -I../iterater -I../matlabio3 -I../matlabio
+//mex -v mesh_base.cpp ../matlabio3/matlabio.cpp -DMIOFLAG=matlab  -I../iterator -I../matlabio3 -I../matlabio
 //[v1,f1,v2,f2]=mesh_base(1,2,1,2,1,2,1,1,1) ;
```
willGraham01, line 769: Can we delete these lines if they are commented out anyway?

tdms/src/numerical_derivative.cpp
```diff
@@ -75,36 +63,25 @@ void init_diff_shift_op( double delta, fftw_complex *Dk, int N){
...
- */
-
 void first_derivative( fftw_complex *in_pb_pf, fftw_complex *out_pb_pf,
 		       fftw_complex *Dk, int N, fftw_plan pf, fftw_plan pb){
   
-  //fprintf(stderr,"01\n");
+  /*
+   *  1. Fourier transform the data in in_pb_pf, placing the result into out_pb_pf
+   *  2. Multiply the result of the FT (out_pb_pf) by the Dk coefficients place
+   *     the result into in_pb_pf.
```
willGraham01, line 72:
```suggestion
   *  2. Multiply the result of the FT (out_pb_pf) by the Dk coefficients, placing
   *     the result into in_pb_pf.
```

willGraham01 (commented):
Looks good, only a couple of minor typos I came across. Have commented on the functionality of parts of the codebase to provide some more context.

tdms/include/numerical_derivative.h
samcunliffe: Ah. You're right. What I _wrote_ is "every element by every element" which would mean some kind of O(n²) operation but we're just mapping complex multiplication across a list.

tdms/include/numerical_derivative.h
samcunliffe: The improvements are anyway clarifying.

tdms/src/mesh_base.cpp
samcunliffe: 🤷 I would leave them in for now... but when we've refactored, documented everything, and are all a bit happier with everything, we can probably nuke all of these commented-out lines.

tdms/include/mesh_base.h
samcunliffe: Do be fair, some documentation is in the source and gets moved/tidied/improved as part of Tom's refactoring. So we can just leave the TODO for now (to avoid annoying him with conflicts).

```
Generate an oriented mesh on the surface of a cuboid
within the FDTD grid.
 ```

File: tdms/include/openandorder.h (hunk 6 above)
Comment (samcunliffe) [reply tdms/include/openandorder.h:51]: Compares what is read in to what is provided to the main call to `openandorder` (the function).
Actually, it doesn't just compare, it assigns the pointers.

File: tdms/include/openandorder.h [hunk 9]
```diff
@@ -1,12 +1,51 @@
...
+ */
 void check_files_can_be_accessed(ArgumentNamespace &args);
 
+/**
+ * @brief Iterate through the matrix names and assign the pointer to each matrix
+ * into the appropriate entry of pointers
+ * 
+ * @param expected 
+ * @param actual 
+ * @param pointers
```
samcunliffe, line 49:
```suggestion
 * @brief Iterate through the matrix names and assign the pointer to each matrix
 * into the appropriate entry of pointers
 * 
 * @param expected The names of the matrices in the MATLAB file.
 * @param actual The actual matrices in the MATLAB file.
 * @param pointers Pointers to the matrices.
```

PR #94, opened 2022-09-16
Fix to steady state mode

prmunro:
I updated the code so that steady state mode is now correct. I picked up a bug introduced during recent refactoring work, whereby there are now two variables storing dt, in particular:

params.dt
dt

both are used. This becomes a problem since in some instances dt can be changed within the code. After refactoring, params.dt was not updated, leading to incorrect behaviour. I will add this as an issue.

### DIFF
--- a/tdms/src/iterator.cpp
+++ b/tdms/src/iterator.cpp
@@ -2508,6 +2508,7 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[]){
     dt_old = dt[0];
     Nsteps_tmp = ceil(2.*dcpi/omega_an[0]/dt[0]*3);
     dt[0] = 2.*dcpi/omega_an[0]*3/Nsteps_tmp;
+    params.dt = dt[0];
   }
   
   //fprintf(stderr,"Pre 16\n");
@@ -5761,12 +5762,17 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[]){
 
     if( exphasorssurface || exphasorsvolume || exdetintegral || (nvertices > 0) ){
       if(sourcemode==sm_steadystate){
-        E.add_to_angular_norm(fte, tind, Nsteps, params);
-        H.add_to_angular_norm(fth, tind, Nsteps, params);
+	if( (tind % Nsteps)==0 ){
+	  E.angular_norm = 0.0;
+	  H.angular_norm = 0.0;
+	}
+	
+        E.add_to_angular_norm(fte, tind % Nsteps, Nsteps, params);
+        H.add_to_angular_norm(fth, tind % Nsteps, Nsteps, params);
 
         for(int ifx=0;ifx<N_f_ex_vec;ifx++){
-          extractPhasorENorm(&E_norm[ifx], fte, tind, f_ex_vec[ifx]*2*dcpi, *dt, Nsteps);
-          extractPhasorHNorm(&H_norm[ifx], fth, tind, f_ex_vec[ifx]*2*dcpi, *dt, Nsteps);
+          extractPhasorENorm(&E_norm[ifx], fte, tind % Nsteps, f_ex_vec[ifx]*2*dcpi, *dt, Nsteps);
+          extractPhasorHNorm(&H_norm[ifx], fth, tind % Nsteps, f_ex_vec[ifx]*2*dcpi, *dt, Nsteps);
         }
       }
       else{
@@ -5978,7 +5984,6 @@ if(runmode == rm_complete && (nvertices>0) )
       }
     }
   }
-  
   //fprintf(stderr,"Pos 15\n");
   //noe set the output
   ndims   = 2;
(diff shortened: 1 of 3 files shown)


tdms/src/iterator.cpp
```diff
@@ -5761,12 +5762,17 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[]){
 
     if( exphasorssurface || exphasorsvolume || exdetintegral || (nvertices > 0) ){
       if(sourcemode==sm_steadystate){
-        E.add_to_angular_norm(fte, tind, Nsteps, params);
-        H.add_to_angular_norm(fth, tind, Nsteps, params);
+	if( (tind % Nsteps)==0 ){
+	  E.angular_norm = 0.0;
+	  H.angular_norm = 0.0;
```
willGraham01, line 5767: Can't we just set these as the default initial values in the `field` class (or the respective subclasses)?
Might as well save an `if` statement here.

tdms/src/iterator.cpp
```diff
@@ -5761,12 +5762,17 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[]){
...
+	}
+	
+        E.add_to_angular_norm(fte, tind % Nsteps, Nsteps, params);
+        H.add_to_angular_norm(fth, tind % Nsteps, Nsteps, params);
 
         for(int ifx=0;ifx<N_f_ex_vec;ifx++){
-          extractPhasorENorm(&E_norm[ifx], fte, tind, f_ex_vec[ifx]*2*dcpi, *dt, Nsteps);
-          extractPhasorHNorm(&H_norm[ifx], fth, tind, f_ex_vec[ifx]*2*dcpi, *dt, Nsteps);
+          extractPhasorENorm(&E_norm[ifx], fte, tind % Nsteps, f_ex_vec[ifx]*2*dcpi, *dt, Nsteps);
+          extractPhasorHNorm(&H_norm[ifx], fth, tind % Nsteps, f_ex_vec[ifx]*2*dcpi, *dt, Nsteps);
```
willGraham01, line 5775: Passing `tind % Nsteps` is **_not_** the same as passing `tind` here - so this changes the functionality of the code here. I don't think that this is the intended outcome of this computation, but if it is I just want to double check it's doing what you think?

This is because of what's going on in `phasor_norm`, which is called by `add_to_angular_norm`. On [this line](https://github.com/UCL/TDMS/blob/1f4207e25134538aeab445a5dfa21480655fab57/tdms/src/electric_field.cpp#L14) we take the exponent. But we don't multiply `(n+1)` by $2\pi$, only take the modulus of the product `omega*dt*(n+1)` against $2\pi$. This means that replacing `n = tind` with `n = tind % Nsteps` doesn't produce the same value.

Concrete example:

Fix `omega` ( $\omega$ ) and `dt` ( $\mathrm{d}_t$ ) at some values, and `Nt = 7` ( $N_t = 7$ ). For the new code to produce an identical output to the old code, we are claiming that `phasor_norm(f, 5, omega, dt, 7)` and `phasor_norm(f, 12, omega, dt, 7)` should produce the same value. However this is only true if 
$$6\omega\mathrm{d}_t \ \mathrm{mod} \ 2\pi = 13\omega\mathrm{d}_t \ \mathrm{mod} \ 2\pi, \Leftrightarrow 7\omega\mathrm{d}_t = 0 \ \mathrm{mod} \ 2\pi,$$
which is (almost surely) _not_ true in general for any pair of values $\omega$, $\mathrm{d}_t$?

If there's some particular properties of the `params` (`omega` and `dt`) that makes the `tind % Nsteps` substitution safe, then this would be OK of course, but I don't think there is?

willGraham01 (changes requested):

tdms/src/iterator.cpp
prmunro: This is a good point, there is something wrong here. I'm working on it.

samcunliffe: The test of files in `arc_08.zip` fails on this branch. As in, running `tdms in/pstd_fs.mat test_out.mat`, the `test_out.mat` doesn't match the reference output file in `out/pstd_fs.mat`.

Not sure if this helps you debug, or is just noise.

prmunro: There was likely an error in the arc_08 files as I think I used an old version of the code the generate the output files. The attached will fix this problem.
[arc_08.zip](https://github.com/UCL/TDMS/files/9632661/arc_08.zip)

willGraham01: > I have gone over this code again in detail and found that the following are equivalent:
> 
> extractPhasorENorm(&E_norm[ifx], fte, tind % Nsteps, f_ex_vec[ifx]_2_dcpi, *dt, Nsteps); extractPhasorENorm(&E_norm[ifx], fte, tind , f_ex_vec[ifx]_2_dcpi, *dt, Nsteps);

Thanks for looking into this Peter, and for adding the comment in the code.

willGraham01 (approved):

tdms/src/iterator.cpp
willGraham01: Yes sorry - forgot to change that :sweat_smile: . Should be ready for merge now!

PR #98, opened 2022-09-20
Two more system tests.

samcunliffe:
This PR is adds two more test cases provided by @prmunro (thanks!) and another 2/4 of #30: a test w/ exi present and a test w/ exdetintegral.

The first is very fast:
```
py.test tdms/tests/system/test_arc10.py  73.70s user 10.40s system 563% cpu 14.916 total
```

The second a little slower (but not the slowest):
```
py.test tdms/tests/system/test_arc09.py  704.33s user 40.89s system 681% cpu 1:49.37 total
py.test tdms/tests/system/test_arc03.py  1171.14s user 66.51s system 527% cpu 3:54.44 total
```
And notably I had to tweak the HDF5File class to cope with groups.

### DIFF
--- a/tdms/tests/system/utils.py
+++ b/tdms/tests/system/utils.py
@@ -36,7 +36,20 @@ class HDF5File(dict):
         super().__init__()
 
         with h5py.File(filepath, "r") as file:
-            self.update({k: self.to_numpy_array(v) for k, v in file.items()})
+            self.update({k: self.to_numpy_array(v) for k, v in self.traverse(file)})
+
+    def traverse(self, file_or_group, prefix=""):
+        """
+        Traverse the hdf5 file, when a group is encountered also traverse the
+        group (get all datasets).
+        """
+        for key in file_or_group.keys():
+            item = file_or_group[key]
+            path = f"{prefix}/{key}" if prefix else key
+            if isinstance(item, h5py.Dataset):
+                yield (path, item)
+            elif isinstance(item, h5py.Group):
+                yield from self.traverse(item, path)
 
     def to_numpy_array(self, dataset: h5py.Dataset) -> np.ndarray:
         """Convert a hdf5 dataset into a numpy array"""
@@ -67,7 +80,7 @@ class HDF5File(dict):
         arrays
         """
 
-        for key, value in self.items():
+        for key, value in self.traverse(self):
 
             if key not in other:
                 return False  # Key did not match
--- a/tdms/tests/system/utils.py
+++ b/tdms/tests/system/utils.py
@@ -8,7 +8,7 @@ from functools import wraps
 from pathlib import Path
 from platform import system
 from subprocess import PIPE, Popen
-from typing import Generator, Union
+from typing import Generator, Tuple, Union
 from urllib import request
 from zipfile import ZipFile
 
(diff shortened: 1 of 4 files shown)


t-young31:
apologies – don't really have time for a proper review today. but in general more tests = better so generally happy. only thought is to reduce the system size so we can be more comprehensive i.e. have both 3D and 2D examples, for instance(?)

samcunliffe:
> apologies – don't really have time for a proper review today. but in general more tests = better so generally happy. only thought is to reduce the system size so we can be more comprehensive i.e. have both 3D and 2D examples, for instance(?)

I like it. Although perhaps we merge this now (since slow protection >> no protection). And I'll ticket myself to speedup the rest of the tests.

tdms/tests/system/utils.py
```diff
@@ -36,7 +36,20 @@ def __init__(self, filepath: Union[str, Path]):
         super().__init__()
 
         with h5py.File(filepath, "r") as file:
-            self.update({k: self.to_numpy_array(v) for k, v in file.items()})
+            self.update({k: self.to_numpy_array(v) for k, v in self.traverse(file)})
+
+    def traverse(self, file_or_group, prefix=""):
```
dstansby, line 41: Add typing to this new method?

dstansby (commented):
Putting each test in a new file is resulting in a lot of boilerplate code - it looks like you're running through the same steps each time, so I'd recommend writing one test function an [parametrizing it](https://docs.pytest.org/en/stable/example/parametrize.html). Something like:
```python
@pytest.mark.parametrize('input_file, reference_file', [('arc_09/in/pstd_fs.mat', 'arc_09/out/pstd_fs.mat'), ...])
def test_tdms():
    ...
```

tdms/tests/system/utils.py
samcunliffe: I didn't because it was going to be something epic like...
```py
def traverse(self, file_or_group: Union[h5py.File, h5py.Group], prefix: str = "") -> Generator[str, h5py.Dataset]:
```
I'm perfectly happy to do this. It's just... a lot.

dstansby: I think that's not too bad, and would be +1 to adding it.

tdms/tests/system/utils.py
```diff
@@ -67,7 +80,7 @@ def matches(self, other: "HDF5File", rtol=1e-10) -> bool:
         arrays
         """
 
-        for key, value in self.items():
+        for key, value in self.traverse(self):
```
t-young31, line 83: picky and annoying I know, and I don't have a good rational, but I don't like the pattern

```python
self.method(self)
```

I may be inclined to do `_traverse(self)`, but also 🤷🏼

t-young31 (approved):
Looks great to me – agree some slow protection is definitely better than no fast protection. However, I'd be inclined to make the tests fast ASAP, to limit the trees we're burning 🌲 🔥

prmunro:
Hi All, I can further shorten the time for tests to complete without compromising the validity of the test. This would mean uploading new test files. Should I do this?

samcunliffe:
> Hi All, I can further shorten the time for tests to complete without compromising the validity of the test. This would mean uploading new test files. Should I do this?

Hey @prmunro , 
Yes, that would be very helpful indeed! We should be able to update the zenodo version and keep the same links and everything. So we potentially don't even need to PR.

Issue #91, opened 2022-09-13
Can't build the unit tests executable on linux


samcunliffe:
@prmunro very sensibly followed the instructions and:

* the `build` dir didn't get populated (why?)
* `cmake .. -DCMAKE_BUILT_TYPE=Test` didn't create the `tdms_test` file (nor did `ctest` work).

Very embarrassing. We should double-check the instructions vs. GitHub actions are doing the same thing.

samcunliffe:
Hey @willGraham01 

For the second bullet, I think we just gave bad instructions. 

To build the `tdms_tests` executable / `ctest`...
```
cd build; cmake .. -DBUILD_TESTING=ON
ctest # or ./tdms_tests
```

And to build with +debug output
```
cd build; cmake .. --DCMAKE_BUILT_TYPE=Debug
```
`CMAKE_BUILT_TYPE=Test` doesn't do anything.


-------------


I can't be sure, but I have a [hypothesis for the first problem](https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#i-run-an-out-of-source-build-but-cmake-generates-in-source-anyway-why).

I can break CMake by having a rogue CMakeCache.txt one directory up from the build dir (this happens if one ran `cmake` from the wrong directory previously).
```
[ccaecun@login13:build]$ pwd
/lustre/home/ccaecun/TDMS/tdms/build
[ccaecun@login13:build]$ ls ../CMake*.txt
../CMakeCache.txt  ../CMakeLists.txt
[ccaecun@login13:build]$ cmake ..
-- Build spdlog: 1.10.0
-- Build type: Debug
-- Configuring done
-- Generating done
-- Build files have been written to: /home/ccaecun/TDMS/tdms
[ccaecun@login13:build]$ ls
[ccaecun@login13:build]$ make install
make: *** No rule to make target `install'.  Stop.
[ccaecun@login13:build]$ 
```
If the option exists, `cmake .. --fresh <other options>` will fix this problem. I have `--fresh` on my Mac's CMake (3.24.1) but not on Myriad (3.21.1).


Do we need to add anything to any documentation anywhere? My feeling: not. I claim it's just that: we need to be careful when giving CMake instructions.

willGraham01:
> I can break CMake by having a rogue CMakeCache.txt one directory up from the build dir (this happens if one ran cmake from the wrong directory previously).

[...]

PR #100, opened 2022-09-22
Wrong build type -> CMake error.

samcunliffe:
Sorry for the supertrivial PR. Logically separate from the other changes, though. At least it's easy to review 🙃 

Discussion in #91. We might not want this, and rather rely on the doc being correct.

### DIFF
--- a/tdms/CMakeLists.txt
+++ b/tdms/CMakeLists.txt
@@ -80,6 +80,10 @@ endif()
 if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug"
         OR "${CMAKE_BUILD_TYPE}" STREQUAL "RelWithDebInfo")
     add_definitions(-DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_DEBUG)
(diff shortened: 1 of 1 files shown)


tdms/CMakeLists.txt
```diff
@@ -80,6 +80,10 @@ endif()
 if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug"
         OR "${CMAKE_BUILD_TYPE}" STREQUAL "RelWithDebInfo")
     add_definitions(-DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_DEBUG)
+elseif("${CMAKE_BUILD_TYPE}" STREQUAL "Test")
+    message(WARNING "CMAKE_BUILT_TYPE \"Test\" is not recognised.
+    Did you mean -DBUILD_TESTING=ON ?")
```
giordano, line 85: Perhaps this should error?  A warning can be easily missed.

giordano (approved):
> We might not want this, and rather rely on the doc being correct.

While I prefer correct docs....

0fc

just to make sure docs are correct.

PR #101, opened 2022-09-23
Parametrizing system tests

dstansby:
This PR is to remove a lot of duplicated code currently used in the system tests. This will primarily be achieved by paramerizing the tests, which means running the same test (running TDMS with an input file, checking output against a reference file) with different parameters (different input and reference files).

To add a new system test, there is now a dictionary of test numbers and download URLs that can be easily modified.


dstansby:
This is now working. I relied on the structure of filenames that the first three tests followed:
- an input file named `pstd_input_file.m`
- one or more input data files named `pstd_{run_type}_input.mat`,
- corresponding reference output data files named `pstd_{run_type}_reference_output.mat`.

where run_type is one of `["fs", "cyl", "sph"]`.

The other tests being added in https://github.com/UCL/TDMS/pull/101 seem to follow this structure, but the existing "08" tests do not, with filenames like `pstd_fs_steady_reference_output_fast.mat`.

Could someone advise on which of these is best:
1. The filenames for the 08 tests could be renamed to follow the pattern of the other tests and re-uploaded?
2. The code for running the tests should accept more general filenames? This increases flexibility but also code complexity.

tdms/tests/system/utils.py
```diff
@@ -178,3 +178,22 @@ def download_data(url: str, to: Path) -> None:
...
+    Checks that the output .mat file (with a HDF5 format) contains tensors with
+    relative mean square values within numerical precision of the reference.
+    """
+    output_filename = "pstd_output.mat"
+    run_tdms(input_filename, output_filename)
+
+    reference = HDF5File(reference_output_filename)
+    output = HDF5File(output_filename)
+    try:
+        assert output.matches(reference)
```
samcunliffe, line 197: Should we have a (maybe optional) assert message passed all the way in here? Or do you think that's overkill?

It's just: I see "blah doesn't match reference". And it's not clear from the error if I've broken the steady-state or the extrapolation or what.

dstansby: I've done two things to improve this:
1. Change the `matches` method to `assert_matches`
2. Used numpy's built in `np.testing.assert_equal` to check the data, which will print useful information if the comparison fails somehow

dstansby:
I've updated the testing code to accept arbitrary input files as long as the filname has the format `pstd_{run_type}_input.mat`. The output files then have to have the format `pstd_{run_type}_reference_output.mat`,  but the `run_type` bit can be arbitrary.

samcunliffe:
Closing as overtaken by #202.
(Thanks for teh code @dstansby !)

PR #102, opened 2022-09-27
Fix the tests.

samcunliffe:


giordano (approved):
:rocket: 

For the benefit of future readers, in the last few days we've seen jobs failing because apparently the workflow `matlab-actions/setup-matlab` is now installing a newer version of Matlab than it used to do until last week (R2022a vs R2022b), and we were hardcoding in our workflow files the root directory of Matlab, which is now broken.  This PR should make the workflow files more future-proof

PR #106, opened 2022-09-30
Obey the GIT_SSH option also for catch2.

samcunliffe:
I missed this when testing #92. 

→ We also need to grab catch2 over ssh if the `GIT_SSH` option is set. 

This PR allows us to also build the tests on Myriad. Or if the user just prefers ssh.

### DIFF
--- a/tdms/cmake/targets.cmake
+++ b/tdms/cmake/targets.cmake
@@ -25,7 +25,7 @@ function(test_target)
 
     FetchContent_Declare(
             Catch2
-            GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+	    GIT_REPOSITORY ${GITHUB_PREFIX}catchorg/Catch2.git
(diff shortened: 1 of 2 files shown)


tdms/cmake/targets.cmake
```diff
@@ -25,7 +25,7 @@ function(test_target)
 
     FetchContent_Declare(
             Catch2
-            GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+	    GIT_REPOSITORY ${GITHUB_PREFIX}catchorg/Catch2.git
             GIT_TAG        v3.0.1
```
samcunliffe, line 29:
```suggestion
            GIT_REPOSITORY ${GITHUB_PREFIX}catchorg/Catch2.git
            GIT_TAG        v3.0.1
```

giordano (approved):

PR #108, opened 2022-10-03
Namespace for Global Variables

willGraham01:
Adds namespaces for the mathematical and physical constants that are defined by the TDMS executable. 

This helps avoid conflicts between names and definitions in `globals.h` and local variables using the same naming conventions in `interator.cpp` and `interpolate.cpp`. Other files that have been produced by the iterator refactoring #63 are also updated accordingly.

Changes are:

#### `TDMS_MATH_CONSTANTS` namespace

Namespacing 
- `DCPI` : $\pi$, see [#64 comment](https://github.com/UCL/TDMS/pull/64#issuecomment-1223843945) as to why we don't just use `M_PI`,
- `IMAGINARY_UNIT` : $\mathrm{i}$, the imaginary unit as a `std::complex<double>`.

There were a further two variables defined here; `zero` and `convergence_level`, which are likely numerical tolerance values. These are unused in the code, so have been deleted.

Further to this point, there is also a case to be made for switching `dcpi` to use `M_PI` or something similar from the math libraries.

#### `TDMS_PHYS_CONSTANTS` namespace

Namespacing 
- `EPSILON0` : previously `eo`; $\varepsilon_0$ (free space permittivity), 
- `MU0` : previously `muo`; $\mu_0$ (free space permeability), 
- `LIGHT_V` : previously `light_v`; $c$ (speed of light in vacuum), 
- `Z0` : previously `Zo`; $Z_0$ (impedance of free space). Again a remark here that `Zo` appears to be unused in the codebase.

Beyond renaming variables, and removing unused ones, no changes have been made to the functionality of the codebase.

---
## MacOS fix (depends on #117)

This branch also contains the fix #117 uses to get the MacOS tests running. Delay merging this PR until #117 is successfully merged.


samcunliffe:
[Capital namesoaces plz no](https://google.github.io/styleguide/cppguide.html#Namespaces).

.github/workflows/macos_tests.yml
```diff
@@ -22,7 +22,6 @@ jobs:
       matrix:
         os:
           - macos-12
-          # TODO: can change to macos-latest (but it's currently aliased to macos-11)
```
samcunliffe, line 25: Ditto from #80

tdms/include/globals.h
```diff
@@ -50,19 +50,21 @@ enum RCSType   { parallel , perpendicular };
 //			Mathematical Constants
 // **************************************
 
-const double dcpi  = 3.14159265358979323846;
-
+namespace tdms_math_constants
+{
+const double dcpi = 3.14159265358979323846;                                  // Pi
+const std::complex<double> IMAGINARY_UNIT = std::complex<double>(0.0, 1.0);  // Imaginary unit
+}
```
samcunliffe, line 57: Relavent Google style recommendations:

* [Namespace naming](https://google.github.io/styleguide/cppguide.html#Namespace_Names) (✅)
* [Namespace formatting](https://google.github.io/styleguide/cppguide.html#Namespace_Formatting)
* [const → constexpr](https://google.github.io/styleguide/cppguide.html#Use_of_constexpr)
* [Const naming](https://google.github.io/styleguide/cppguide.html#Constant_Names)

A bit annoying: as a global `IMAGINARY_UNIT` is what I'm used to. But as a const(expr?) in a namespace, it seems GSG prefers `kImaginaryUnit`. I'm not super opinionated on this point, and `readability counts` 🐍 .

I found [this scicomp stackexchange post](https://scicomp.stackexchange.com/a/31003) (there is a scicomp stackexchange!). Certainly I'm more used to what is suggested there:

```c++

namespace tdms_math_constants {

constexpr double PI = 3.14159265358979323846;                                    // Pi
constexpr std::complex<double> IMAGINARY_UNIT = std::complex<double>(0.0, 1.0);  // Imaginary unit

}

//--

namespace tdms_phys_constants {

constexpr double E0 = 8.85400e-12;                             // free space electric permitivity
constexpr double M0 = 4.0 * tdms_math_constants::PI * 1.0e-7;  // free space magnetic permeability
constexpr double LIGHT_V = 1.0 / sqrt(M0 * E0);                // free space light velocity
constexpr double Z0 = 376.734;                                 // free space inpedance

}
```

whereas GSG would have us do:

```c++

namespace tdms_math_constants {

constexpr double kPi = 3.14159265358979323846;                                   // Pi
constexpr std::complex<double> kImaginaryUnit = std::complex<double>(0.0, 1.0);  // Imaginary unit

}

//--

namespace tdms_phys_constants {

constexpr double kE0 = 8.85400e-12;                              // free space electric permitivity
constexpr double kM0 = 4.0 * tdms_math_constants::kPi * 1.0e-7;  // free space magnetic permeability
constexpr double kLightV = 1.0 / sqrt(M0 * E0);                  // free space light velocity
constexpr double kZ0 = 376.734;                                  // free space inpedance

}
```

At this point, I'd simply pick your favourite and be consistent. But I wouldn't mix variable naming styles for the same thing (as you have right now). Let's get it right once and forall now that we're making a noisy "touches many files" -type PR.

tdms/include/globals.h
```diff
@@ -50,19 +50,21 @@ enum RCSType   { parallel , perpendicular };
...
-const double Zo		 = 376.734;						// free space inpedance
-
-const double zero			   = 1.0e-12;			// zero level for numerical comparisons
-const double convergence_level = 1.0e-10;			// zero level for convergence
+namespace tdms_phys_constants
+{
+const double eo = 8.85400e-12;                               // free space electric permitivity
+const double mo = 4.0 * tdms_math_constants::dcpi * 1.0e-7;  // free space magnetic permeability
+const double light_v = 1.0 / sqrt(mo * eo);                  // free space light velocity
+const double Zo = 376.734;                                   // free space inpedance
```
samcunliffe, line 68: As you say, the notmal math convention has these guys as: $\epsilon_0$, $\mu_0$, $c$, and $Z_0$.

Well `0`!= `o` (is `o` chosen because it looks like a sub?). And (a matter of taste, but) would `kEpsilon0`, `kMu0` (or `EPSILON0`, `MU0`) be more readable in your opinion?

Again: if we're doing noisy changes anyway, let's get it right first time.

samcunliffe (approved):
> dcpi : , although I don't quite know why we don't just use M_PI),

On modern systems, [there is no reason not to](https://stackoverflow.com/a/1727896/1444054). And we're C++17 😎 .

I'd say use it, but then there's only one `tdms_mathematical_constant`, so probably doesn't warrant it's own namespace.


***

Approving. Probably worth waiting on #88 's merge and do the find-replace _once_ when merging in `main` to here.

giordano:
> > dcpi : , although I don't quite know why we don't just use M_PI),
>
> On modern systems, [there is no reason not to](https://stackoverflow.com/a/1727896/1444054). And we're C++17 😎 .

For reference, this came up already: https://github.com/UCL/TDMS/pull/64#issuecomment-1223843945

samcunliffe:
> > > dcpi : , although I don't quite know why we don't just use M_PI),
> > 
> > 
> > On modern systems, [there is no reason not to](https://stackoverflow.com/a/1727896/1444054). And we're C++17 😎 .
> 
> For reference, this came up already: [#64 (comment)](https://github.com/UCL/TDMS/pull/64#issuecomment-1223843945)

Ugh. Windows is so uncool.
OK. Let's stick with the historic decision.

Perhaps rather than having 3.1415.... we could do `std::atan(1)*4` ?

tdms/include/globals.h
willGraham01: The mixture is present in the original code - I'll convert everything to SciComp standards because that's what I'm used to seeing too. This PR is going to look a lot bigger after these changes just because of all the name changes...

tdms/include/globals.h
willGraham01: Agreed, I thought it odd when I did my first pass but otherwise accepted it. With going over to SciComp standards, I'll replace these with `0`s.

tdms/tests/unit/test_numerical_derivative.cpp
```diff
@@ -2,10 +2,12 @@
...
  * @brief Tests of the numerical differentiation/FFT functions.
  */
-#include <catch2/catch_test_macros.hpp>
-#include <catch2/catch_approx.hpp>
 #include <fftw3.h>
 #include <spdlog/spdlog.h>
+
+#include <catch2/catch_approx.hpp>
+#include <catch2/catch_test_macros.hpp>
+
```
samcunliffe, line 10: Are these headers moving because of the new clang-format dotfile?

samcunliffe (approved):

tdms/tests/unit/test_numerical_derivative.cpp
willGraham01: I imagine so - I didn't touch them but VSCode is meant to do some auto-formatting on file close.
I thought
```c++
#include <foo.bar>
```
without a space after the hash was the usual standard?

samcunliffe: > without a space after the hash was the usual standard?

Yeah. GSG doesn't have spaces.

PR #119, opened 2022-10-04
Fix the macOS build and add concurrency.

samcunliffe:
The underlying issue is an update ~~of Xcode used in~~ the runner images: actions/runner-images/issues/6225
This is neater and targets main (#117 is to Will's branch).

~~🚧 testing the concurrency to save 🌳🔨~~
Concurrency is working ✅

### DIFF
--- a/.github/workflows/macos_tests.yml
+++ b/.github/workflows/macos_tests.yml
@@ -38,7 +38,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          brew install fftw libomp python3
+          brew install fftw llvm python3
 
       - name: Install Python dependencies
         shell: bash
@@ -58,7 +58,7 @@ jobs:
           cmake ${GITHUB_WORKSPACE}/tdms -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON\
            -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}\
            -DBUILD_TESTING=${{ matrix.build_testing }}\
-           -DOMP_ROOT=/usr/local/opt/llvm/ -DCXX_ROOT=/usr/local/opt/llvm
+           -DOMP_ROOT=/usr/local/opt/llvm/ \
 
       - name: Build TDMS
         working-directory: ${{runner.workspace}}/build
--- a/.github/workflows/macos_tests.yml
+++ b/.github/workflows/macos_tests.yml
@@ -7,6 +7,12 @@ on:
   release:
   workflow_dispatch:
 
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: always.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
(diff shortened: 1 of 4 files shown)


.github/workflows/macos_tests.yml
```diff
@@ -38,7 +44,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          brew install fftw libomp python3
+          brew install fftw llvm python3
```
giordano, line 47: Is installing llvm now necessary? What's the problem with new apple clang?

giordano (commented):
Looks good, but it isn't clear to me what's exactly the problem with apple clang

.github/workflows/macos_tests.yml
samcunliffe: Yeah. So my current understanding: 
* `llvm` was not (and still is not) necessary for Xcode 13. Because it's included.
* `llvm` is not included in Xcode 14 (‽) <sup>[Citation very much needed]</sup>. I *have* confirmed that it doesn't work without.
* `libomp` was never needed on it's own because it's included with `llvm` libraries.
* No need to point CMake to LIBCXX because it can find it by itself. And that path was wrong for Xcode 14.

Regarding the last point, it might  *never* have been needed to be pointed to LIBCXX on these runners. Just necessary on M1/2. And I from which I copied the build procedure from my laptop.

samcunliffe: I would be happy if I'm wrong and this can be simplified even further, but unless I'm missing a trick...

giordano: > llvm is not included in Xcode 14 (‽) [Citation very much needed]. I have confirmed that it doesn't work without.

Yeah, I'm not convinced that's true, the images definitely come with apple clang available out-of-the-box: https://github.com/actions/runner-images/blob/2aec524970584138c5986109ca7be7ae89b0899d/images/macos/macos-12-Readme.md#language-and-runtime. Note also that vanilla clang from homebrew should already be available (but under a different path).

I'd just like to understand what's going on exactly if possible, because if there is something wrong with apple clang we should either fix it or note it in the documentation.

samcunliffe: On my dev branch:
* 4f8d351 [fails, can't find OpenMP](https://github.com/UCL/TDMS/actions/runs/3190080975)
* c52b4b8 [also fails, can't find OpenMP](https://github.com/UCL/TDMS/actions/runs/3190153934)

samcunliffe: Hmm. OK so it seems I can use the already-installed llvm's libomp via:
```
$(brew --prefix llvm@14)/lib
```

But this is hard-coding a "14". I think I'd rather run
```
brew install llvm
```
and use either same line without specifying `@14`.

samcunliffe: It's also consistent with the Mac-specific instructions.

willGraham01 (approved):

Issue #115, opened 2022-10-04
Remove now-redundant functions (and headers).


willGraham01:

samcunliffe:
* #134

PR #131, opened 2022-10-13
Replace interpolation functions

### LINKED ISSUE #113: Change the interpolation calls in `exphasorssurface`, `exphasorsvolume`, and `fieldsample` to the new interpolation functions.

willGraham01:
Closes #113 | Depends on #130 |


willGraham01:
Also removes (with the objective of making #115 easier):
- `interpolate.cpp`, `interpolate.h`
- `interp{1,2,3}` - old cubic interpolation functions
- References to these files in `CMakeLists.txt`

There are still methods for `TE` and `TM` interpolation - these should be deleted awaiting answer from #113.

willGraham01:
Closing due to conflicts with reworking interpolation functionality

Issue #129, opened 2022-10-10
Memory leak in `TDFieldExporter2D::export_field`


samcunliffe:
Not sure why yet, but we have a segmentation violation (I think from the 2D field exporter)...
```
➤ tdms in_pstd_cyl.mat output.mat
[2022-10-10 16:47:51.417] [debug] Parsing 3 command line arguments
infile:[in_pstd_cyl.mat], outfile:[output.mat], m=0
[2022-10-10 16:47:51.417] [debug] Finished parsing arguments
[2022-10-10 16:47:51.425] [info] Using 8 OMP threads
[2022-10-10 16:47:51.425] [info] Isource is empty
[2022-10-10 16:47:51.425] [info] Jsource is empty
[2022-10-10 16:47:51.425] [debug] f_ex_vec has ndims=2 N=1
[2022-10-10 16:47:51.425] [info] interpolation method = 1
[2022-10-10 16:47:51.425] [info] exi not present
[2022-10-10 16:47:51.425] [info] eyi not present
[2022-10-10 16:47:51.425] [info] Ni_tdf=276 Nk_tdf=276
[2022-10-10 16:47:51.425] [info] Refractive index = 1.35001
[2022-10-10 16:47:52.470] [debug] init_diff_shift_op, delta=-0.5
[2022-10-10 16:47:52.470] [debug] init_diff_shift_op, delta=-0.5
[2022-10-10 16:47:52.470] [debug] init_diff_shift_op, delta=-0.5
[2022-10-10 16:47:52.470] [debug] init_diff_shift_op, delta=0.5
[2022-10-10 16:47:52.470] [debug] init_diff_shift_op, delta=0.5
[2022-10-10 16:47:52.470] [debug] init_diff_shift_op, delta=0.5
[2022-10-10 16:47:52.470] [info] Np=2 Nt=500 Npe=250 f_max=2.94577e+14 Npraw=0.4606
dims:(254,1,254)
[2022-10-10 16:47:52.476] [debug] Starting main loop
[2022-10-10 16:47:52.488] [info] Saving field
[2022-10-10 16:47:52.488] [debug] Entered TDFieldExporter2D::export_field
[2022-10-10 16:47:52.488] [debug] Entered field written to array, creating output filename
[2022-10-10 16:47:52.488] [debug] output folder name: ??
[2022-10-10 16:47:52.488] [debug] output file name: ??/ex_000000.mat
time domain output: ??/ex_000000.mat
[...]

PR #133, opened 2022-10-14
C++ 14 → 17.

samcunliffe:
Going via a one-liner PR to separate out any problems we might have from **_this_** change, vs my eventual fix to #129 which --I think-- will need some `std::filesystem` stuff.

At least trivial PRs are easy to review 😛. 

If you approve: feel free to just merge in at whatever point seems sensible in your merge chain #130 → #80 → #131(?)

### DIFF
--- a/tdms/CMakeLists.txt
+++ b/tdms/CMakeLists.txt
@@ -2,7 +2,7 @@
 cmake_minimum_required(VERSION 3.21)
 
 project(tdms LANGUAGES CXX)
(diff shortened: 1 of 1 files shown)


samcunliffe:
Let me also add @prmunro just for sanity.
@prmunro are you OK with this? Are you using gcc 8.x ? Seems [most features are available since gcc 7](https://gcc.gnu.org/projects/cxx-status.html#cxx17) but not quite all.

willGraham01 (approved):

Issue #132, opened 2022-10-14
Remove `int i, int j, int k` data-clump


willGraham01:
Yee cell indices are referenced by three indices. Such indices are often passed to functions/methods as three separate inputs, however no two indices make sense without the others - even in the case of a 2D simulation (no y-direction) the `j` index is set to 0 at all times.

A `coordinate` struct (or type) should be declared in `globals.h` with attributes `.i, .j, .k` to reduce the bloat, and the appropriate functions reworked to adopt this new functionality. The majority of these functions are in the `iterator.cpp` file, and concern the interpolation methods for the various `Field`s and `SplitField`s.

In the interest of minimising conflicts, it would be advisable to resolve #115 prior to addressing this issue.

***

### Dependencies

* #115

PR #135, opened 2022-10-18
Yee cell dimensions and coordinate clump class

willGraham01:
Addresses #132 |

Adds a class for storing the indices of Yee cells (i,j,k). This is a class rather than a struct as additional functionality (EG computing the absolute position of the cell centre, or the location of the associated field components) is likely to be added in future.

Interpolation functions will be updated to accept instances of this class following merge of #113.

---
Adds the struct `YeeCellDimensions` to replace the `dx,dy,dz` `doubles` that populate `iterator.cpp`. These values are read in at the beginning of the function, and are simulation-specific, ergo should be an attribute of `SimulationParameters` rather than loose variables. Calls in `iterator` have been updated accordingly.


tdms/include/cell_coordinate.h
```diff
@@ -0,0 +1,28 @@
...
+  };
+  CellCoordinate(int *indices, int buffer_start = 0) {
+    cell_i = indices[buffer_start];
+    cell_j = indices[buffer_start + 1];
+    cell_k = indices[buffer_start + 2];
+  };
+
+  int i() { return cell_i; };   // fetch i
+  int j() { return cell_j; };   // fetch j
+  int k() { return cell_k; };   // fetch k
```
samcunliffe, line 27:
```suggestion
  int i() { return cell_i; };   //< fetch i
  int j() { return cell_j; };   //< fetch j
  int k() { return cell_k; };   //< fetch k
```
(for doxygen, otherwise `int j()` gets documented with "fetch i")

samcunliffe (approved):
👍

Issue #123, opened 2022-10-05
Clean up/simplify build after MATLAB removal


samcunliffe:
>        Worth breaking this out into another issue, also mentioning that we should turn this option from a compile-time option to a command-line option.

_Originally posted by @willGraham01 in https://github.com/UCL/TDMS/pull/122#discussion_r987908780_

PR #122, opened 2022-10-05
Add FDTD test.

samcunliffe:
Adds an example FDTD test (current functionality). Not necessarily meaningful, but a run of the example scripts with FD compile flag. This at least tests against regression.

Also runs the unit tests with the FD flag in the CI for BUILD_TESTING builds. And update the documentation.

***
### Dependencies

- #121 
- #119

PR #138, opened 2022-10-18
Rename mexFunction.

samcunliffe:
Baby step towards:
* #70


samcunliffe:
If this is all fine, I can revert/simplify the complexity introduced in #122 (which would solve #123 even before MATLAB is fully removed).

samcunliffe:
Close in favour of #139.

Issue #89, opened 2022-09-12
Test of parallelisation


samcunliffe:
_Originally posted by @t-young31 in https://github.com/UCL/TDMS/pull/85#discussion_r967261269_

We build and run tests on all three major OS. Perhaps separate out the system tests. And perhaps test parallelisation.

samcunliffe:
Closed by #234

PR #139, opened 2022-10-19
Compiler flag FDFLAG → command-line option.

samcunliffe:
This PR is a first step towards removing MATLAB (#70).

* Renames mexFunction. (3f3ab5ae6e2a21aad636000edbceb116e5fdaa09)
* Removes the FDFLAG compiler flag and replaces it with a command-line option.
* Reverts the changes to build testing introduced in (#122) and solves:
      *  #123.

It obviously touches `iterator.cpp` in a lot of places (and changes the alignment). Note that I don't attempt any kind of factorisation, tidying, or anything like that. It's just `#ifdef`/`#ifndef` to normal if statements.

***

A couple of thoughts for reviewers' consideration:

* @willGraham01, I think you understand better than me, but it's not strictly a `DerivativeType`? Open to better naming suggestions.
* Any better name for the `mexFunction`? I was a bit flippant with `moaf`. But now I can't think of a better name.
* Should we block this PR with: #89?

***

@t-young31 if you have time, your 👀 would be very appreciated. Although... you probably don't have time.


samcunliffe:
The failing builds are an echo of the old CI yml files that haven't been recognised as updated yet... unsure why.

.github/workflows/linux_tests.yml
```diff
@@ -83,19 +79,9 @@ jobs:
...
-
-      - name: Run FDTD TDMS system tests
-        if: matrix.derivative_type == 'FD'
-        working-directory: ${{runner.workspace}}/build
-        shell: bash
-        run: |
-          export OMP_NUM_THREADS=1
-          pytest ${GITHUB_WORKSPACE}/tdms/tests/system/ -s -x  -m fdtd_build_only
\ No newline at end of file
+          pytest ${GITHUB_WORKSPACE}/tdms/tests/system/ -s -x
```
samcunliffe, line 87:
```suggestion
          pytest ${GITHUB_WORKSPACE}/tdms/tests/system/ -s -x

```

.github/workflows/macos_tests.yml
```diff
@@ -85,21 +81,11 @@ jobs:
...
-
-      - name: Run FDTD TDMS system tests
-        # run system tests in mac only on pushes to main (they are a bit slow)
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' && matrix.derivative_type == 'FD' }}
-        working-directory: ${{runner.workspace}}/build
-        shell: bash
-        run: |
-          export OMP_NUM_THREADS=1
-          pytest ${GITHUB_WORKSPACE}/tdms/tests/system/ -s -x  -m fdtd_build_only
+          pytest ${GITHUB_WORKSPACE}/tdms/tests/system/ -s -x
```
samcunliffe, line 91:
```suggestion
          pytest ${GITHUB_WORKSPACE}/tdms/tests/system/ -s -x

```

tdms/tests/system/test_fdtd.py
```diff
@@ -26,13 +24,14 @@ def test_fs():
     """
 
     # misleading filename here
-    run_tdms("arc_example_fdtd/in_pstd_fs.mat", "fdtd_fs_output.mat")
+    run_tdms(
```
giordano, line 27: Do I understand correctly that this is now always running the tests with finite differences only?

giordano (commented):
Oooh, this is lovely!

tdms/include/iterator.h
```diff
@@ -13,6 +13,12 @@
 #include "field.h"
 #include "grid_labels.h"
 
+/**
+ * @brief The Mother of all Functions
+ * Used to be the MATLAB mexFunction
+ */
+void moaf(int, mxArray**, int, const mxArray**, DerivativeMethod);
```
willGraham01, line 20: I guess "execute_simulation" would be a reasonable name? It depends what we do as we refactor `iterator.cpp` #63 to be honest. 

My vote is still for "Lord of all functions (loaf)".

tdms/src/argument_parser.cpp
```diff
@@ -51,6 +51,7 @@ void ArgumentParser::print_help_message(){
                  "openandorder [options] infile gridfile outfile\n"
                  "Options:\n"
                  "-h:\tDisplay this help message\n"
+                 "--finite-difference:\tUse finite-difference derivatives\n"
```
willGraham01, line 54:
```suggestion
                 "--finite-difference:\tUse finite-difference solver, instead of pseudo-spectral method.\n"
```

tdms/tests/system/test_fdtd.py
samcunliffe: Yes, exactly. The finite difference code results in a different reference output. So it never makes sense to run over this file without nor the others with.

.github/workflows/macos_tests.yml
```diff
@@ -28,10 +28,6 @@ jobs:
         build_testing:
           - ON
           - OFF
-        # TODO: remove this once MATLAB dependency is removed and mexFunction is replaced
```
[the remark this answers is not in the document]
samcunliffe: I'll have a go at merging them into a single CI.yml and see how messy it looks...

* #141

tdms/src/openandorder.cpp
```diff
@@ -60,8 +60,13 @@ int main(int nargs, char *argv[]){
     openandorder(args.grid_filename(), (char **)matrixnames_gridfile, matrixptrs, 1);
   }
 
+  // decide which derivative method to use (PSTD or FDTD)
+  DerivativeMethod method = PseudoSpectral; // default
```
willGraham01, line 64:
```suggestion
  SolverMethod method = PseudoSpectral; // default
```
I think `SolverMethod` is a better description, since my understanding is that psuedo-spectral and finite-difference refer to the method employed when time-evolving the system. I (think) I've caught all the instances of `DerivativeMethod` in my review, but I haven't touched `iterator.cpp` so a find+replace is be needed there. 

Tagging @prmunro in case I'm misunderstanding. Finite-difference timestepping happens in the the time-domain, based off the usual difference equation $$f(t_{i+1}) = \Delta_t f'(t_{i}) + f(t_{i}).$$
Psuedo-spectral methods do the timestepping in frequency (Fourier) space instead, since differentiation in real/time corresponds to multiplication in Fourier/frequency space:
$$\frac{\mathrm{d}f}{\mathrm{d}t}(t+\Delta_{t}) = \mathcal{F}^{-1}[D_t\mathcal{F}[f(t)]],$$
where the $D_t$ is the operator that `init_diff_shift_op` computes.

tdms/src/openandorder.cpp
```diff
@@ -60,8 +60,13 @@ int main(int nargs, char *argv[]){
...
+  if (args.finite_difference()) 
+    method = DerivativeMethod::FiniteDifference;
```
willGraham01, line 66:
```suggestion
    method = SolverMethod::FiniteDifference;
```

tdms/include/globals.h
```diff
@@ -50,8 +50,9 @@ enum AxialDirection
 //  Enumerated constants
 // **********************
 
-enum ModeOfRun { Pass1 , Pass2 };
-enum RCSType   { parallel , perpendicular };
+enum ModeOfRun   	  { Pass1 , Pass2 };
+enum RCSType     	  { parallel , perpendicular };
+enum DerivativeMethod { PseudoSpectral, FiniteDifference };
```
willGraham01, line 55:
```suggestion
enum SolverMethod { PseudoSpectral, FiniteDifference };

```

tdms/include/iterator.h
willGraham01, line 20:
```suggestion
void moaf(int, mxArray**, int, const mxArray**, SolverMethod);
```

tdms/src/openandorder.cpp
t-young31: `DerivativeMethod`  is my shoddy naming – apologies

tdms/include/iterator.h
samcunliffe: I went with `execute_simulation`.

t-young31 (approved):
only skimmed but looks great – thanks 😄

tdms/include/argument_parser.h
```diff
@@ -62,6 +64,15 @@ class ArgumentNamespace{
      */
     static bool is_a_flag_argument(std::string arg);
 
+    /**
+     * @brief Check whether we were asked to use finite-difference derivatives.
```
samcunliffe, line 68:
```suggestion
     * @brief Check whether we were asked to use the finite-difference method.
```

tdms/include/iterator.h
```diff
@@ -13,6 +13,12 @@
 #include "field.h"
 #include "grid_labels.h"
 
+/**
+ * @brief The Mother of all Functions
```
samcunliffe, line 17:
```suggestion
 * @brief Executes the main simulation.
```

samcunliffe:
I'm going to hold off on merging until I've done #140.

willGraham01 (approved):
Approving, although holding off merging due to the latest MacOS `omp` [issue causing the build to fail](https://github.com/UCL/TDMS/actions/runs/3295655886/jobs/5434423814).

Fairly sure it's safe to merge this in though, but will hold off for the time being.

PR #143, opened 2022-10-21
Replace Cubic with Band Limited Interpolation

### LINKED ISSUE #113: Change the interpolation calls in `exphasorssurface`, `exphasorsvolume`, and `fieldsample` to the new interpolation functions.

### LINKED ISSUE #115: Remove now-redundant functions (and headers).

willGraham01:
Closes #113 | Closes #115 - at least in the interpolation context, which was the original context of the issue, as #44 tracks it.

### Replaces Interpolation Functions

The old cubic interpolation methods are replaced with their band-limited interpolation equivalents, which are now function methods for the various `Field` and `SplitField` objects.

#### Impacts on Test Performance and Results

System tests are expected to fail due to the reference-runs they were generated from using the old cubic interpolation schemes. A sample visual comparison of the outputs (`arc_01-cyl`) is in the image below.
old-vs-new-arc_01-cyl

Below is a sample from various system tests of the "most different" (in terms of greatest mean-square-difference) outputs from their corresponding reference. These MSDs are of the order $10^{-1}$ in the worst case. The general shape of the computed fields remains the same, with slight numerical discrepancies.
MSD-Worst-Compare1

There are also MATLAB files that can produce these plots for any output file that is produced - these can be added to the repository if desired (although these are results-parsing, rather than central to the TDMS executable's functionality).

#### Unused files and cleanups (#115)

The `InterpolationScheme` class and `best_scheme` function supersede the original cubic-interpolation functions. 

- `interpolate.h` and `interpolate.cpp` are no longer needed and have been deleted. 
- `CMakeLists.txt` has been updated accordingly.
- The old cubic methods `interp{1,2,3}` and `CheckInterpolationPoints` have been deleted, as these are no longer required.
- `test_interpolation_functions.cpp` no longer tests the old `interp{1,2,3}` methods.
- `test_interpolation_determination.cpp` no longer runs `CheckInterpolationPoints`, since this has been deleted as it is unused.


samcunliffe:
The bottom right in plots matrix here:



Looks quite different to my eyes. Probably we should flag this to @prmunro (we can get his opinion IRL on Tuesday).

tdms/src/fields/base.cpp
```diff
@@ -57,7 +57,88 @@ complex<double> Field::phasor_norm(double f, int n, double omega, double dt, int
...
+            case THREE:
+              // 3D interpolation is identical for both E and H fields
+              x_at_centre = interpolate_to_centre_of(AxialDirection::X, i, j, k);
+              y_at_centre = interpolate_to_centre_of(AxialDirection::Y, i, j, k);
+              z_at_centre = interpolate_to_centre_of(AxialDirection::Z, i, j, k);
+              break;
+            case TE:
+              interpolate_TE_components(i, j, k, &x_at_centre, &y_at_centre, &z_at_centre);
+              break;
+            case TM:
```
samcunliffe, line 123: I'd love it if you could change the enumerator `TE` and `TM` to be full words too.
(I don't notice it in the diff, so I can't comment in a more reasonable place.)

tdms/src/iterator.cpp
```diff
@@ -4873,86 +4863,73 @@ void extractPhasorsSurface(double **surface_EHr, double **surface_EHi, ElectricS
...
+        //now update the master array
+        surface_EHr[2][vindex] = surface_EHr[2][vindex] + real(subResultE);
+        surface_EHi[2][vindex] = surface_EHi[2][vindex] + imag(subResultE);
 
-      surface_EHr[5][vindex] = surface_EHr[5][vindex] + real(subResultH);
-      surface_EHi[5][vindex] = surface_EHi[5][vindex] + imag(subResultH);
-    }
+        surface_EHr[5][vindex] = surface_EHr[5][vindex] + real(subResultH);
+        surface_EHi[5][vindex] = surface_EHi[5][vindex] + imag(subResultH);
+      }
```
samcunliffe, line 4932: I miss why the indentation changed here.

tdms/tests/unit/test_interpolation_determination.cpp
```diff
@@ -10,97 +10,39 @@
...
-    - j_u > J-2 with text Interpolation error: j_u too large
-    - k_u > K-2 with text Interpolation error: k_u too large
-    When provided input arguments that do not satisfy any of the above - even cases such as i_l > i_u - should pass without throwing an exception */
-    for (int k_l = 2; k_l < K - 2; k_l++)
-        for (int k_u = 2; k_u < K - 2; k_u++)
-            for (int j_l = 2; j_l < J - 2; j_l++)
-                for (int j_u = 2; j_u < J - 2; j_u++)
-                    for (int i_l = 2; i_l < I - 2; i_l++)
-                        for (int i_u = 2; i_u < I - 2; i_u++)
-                            CHECK_NOTHROW(checkInterpolationPoints(i_l, i_u, j_l, j_u, k_l, k_u, I, J, K));
```
samcunliffe, line 70: 😹

samcunliffe (approved):
All looks good on a first pass. And we're going in the right direction with fewer lines and less duplication 👍 .

Do we need to present this to Peter and get his review of the algorithm/physics?

tdms/src/fields/base.cpp
```diff
@@ -38,7 +38,7 @@ Field::Field(int I_total, int J_total, int K_total) {
 
 void Field::allocate() {
   for (auto arr : {&real, &imag}) {
-    arr->allocate(I_tot+1, J_tot+1, K_tot+1);
+    arr->allocate(I_tot, J_tot, K_tot);
```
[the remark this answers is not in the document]
willGraham01: Agreed :cry:

tdms/src/fields/base.cpp
```diff
@@ -57,7 +57,88 @@ complex<double> Field::phasor_norm(double f, int n, double omega, double dt, int
...
+  // prepare output dimensions
+  const int ndims = 3;
+  int outdims[ndims] = {i_upper - i_lower + 1, j_upper - j_lower + 1, 1};
+  if (mode == THREE) {
+    outdims[2] = k_upper - k_lower + 1;
+    if (outdims[1] < 1) {
+      // full simulation (all components being computed) but in 2D - allow one cell in the y-direction to avoid NULL pointers
+      outdims[1] = 1;
+    }
+  }
```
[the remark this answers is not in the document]
willGraham01: This is partially a MATLAB relic and partially because `XYZTensor3D`s members `x`,`y`, and `z` are not `Tensor3D` instances. For some reason, we don't store the array size in the `XYZTensor3D` (although we do in the `Tensor3D` class).

I advocate `x,y,z` to be converted to `Tensor3D` instances in the next refactor too, and `XYZTensor3D` to have a method `fetch_dimensions` or something. `x,y,z` are all designed to be of the same dimensions, so we can just fetch the dimensions of EG `x` when calling this method.

tdms/include/field.h
```diff
@@ -272,10 +272,45 @@ class Field : public Grid{
...
+  void interpolate_over_range(mxArray **x_out, mxArray **y_out, mxArray **z_out,
+                              Dimension mode = Dimension::THREE);
+
+  /**
+   * @brief Interpolates the Field's TE components to the centre of Yee cell i,j,k
+   *
+   * @param[in] i,j,k Yee cell index
+   * @param[out] x_at_centre,y_at_centre,z_at_centre Addresses to write interpolated values for the x,y,z components (respectively)
+   */
+  virtual void interpolate_TE_components(int i, int j, int k, std::complex<double> *x_at_centre,
```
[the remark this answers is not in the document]
willGraham01: @willGraham01 tagging myself as a reminder to go through this in one sweep for consistency

tdms/src/iterator.cpp
willGraham01: It shouldn't have - probably an artefact of me untangling some of the interpolation logic above. It's now indented properly.

prmunro:
> The bottom right in your plots matrix, here:
> 
> 
> 
> Looks quite different to my eyes. Probably we should flag this to @prmunro (we can get his opinion IRL on Tuesday).

Thanks for these, indeed, it looks as though something has gone awry. Could we see Ex also? Which test is this by the way (apologies if that should be obvious).

Peter

willGraham01:
Those two are `arc_03` - the first being the `sph` test and the latter the `fs` test. Below are all of the field components:

`fs` test output:
arc_03-fs-outputs

`sph` test output:
arc_03_sph-outputs

tdms/src/fields/base.cpp
willGraham01: This is now done, although the PR now touches a couple of other files. Shouldn't cause any major problems though.

samcunliffe:
→ #151

Issue #136, opened 2022-10-18
Add/extend unit tests for classes in field.h and arrays.h


samcunliffe:
Field has some tests, but these can probably be extended. Also need to extend to test functionality in arrays.h

This will be important to ensure we don't break anything when replacing MATLAB.
In particular: allocation and deallocation of instances of these classes.

### Basic Types
- [x] #152
- [x] #153
- [x] [`MaterialCollection`](https://github.com/UCL/TDMS/blob/0f00b6ecb53c81ac8cad1f396b77852c971dbc48/tdms/include/arrays.h#L89-L92) -> Doesn't actually have any attributes, and only has _one_ protected method `init_xyz_vectors`. Ergo, doesn't have it's own test file.
- [x] #158
- [x] #154
- [x] #155
- [x] #156

### Classes that have base classes as attributes
- [x] [`CCollectionBase`](https://github.com/UCL/TDMS/blob/0f00b6ecb53c81ac8cad1f396b77852c971dbc48/tdms/include/arrays.h#L94-L99) -> Doesn't have a constructor or any methods, and just stores three `XYZVectors` as attributes.
- [x] [`DCollectionBase`](https://github.com/UCL/TDMS/blob/0f00b6ecb53c81ac8cad1f396b77852c971dbc48/tdms/include/arrays.h#L119-L123) -> Doesn't have a constructor or any methods, and just stores two `XYZVectors` as attributes.
- [x] #161
- [x] #162
- [x] #163
- [x] #166
- [x] #173 
- [x] #167

### Subclasses Types
- [x] #178
- [x] #184
- [x] #179 
- [x] #185
- [x] #174
- [x] #177
- [x] #175
- [x] #172 
- [x] #170
- [x] `CCoefficientMatrix` (`Matrix<double>`) - No additional functionality on top of the base class.
- [x] #176

### Test classes for cleanup

- [ ] #165 

## `Field.h` classes

- [x] Expand existing tests, add Docstrings where possible, add unit tests that are missing

PR #148, opened 2022-10-27
Basic tests for `arrays.h` datatypes

### LINKED ISSUE #152: `XYZTensor3D`

### LINKED ISSUE #153: `XYZVectors`

### LINKED ISSUE #154: `Matrix`

### LINKED ISSUE #155: `Vector`

### LINKED ISSUE #156: `Tensor3D`

willGraham01:
Addresses #136 |

Adds the `array_tests` folder, containing unit (integration?) tests for the `Tensor3D` and `XYZTensor3D` classes, which will form the backbone of the codebase upon removal of MATLAB.

This change also adds some additional templating to these classes, to make the future transition from MATLAB easier. In particular, `XYZTensor3D` is now templated, so the `Field` class (and its `ElectricField` and `MagneticField` subclasses) no longer need `real` and `imag` parts stored separately - one can simply use one `XYZTensor<complex<double>>` object instead. This change has not been implemented in this PR since this would be noisy and touch a lot of files.

## Other tests that people think would be valuable are welcome!

The main focus of this PR is tests for all the "base" classes in `field.h` and `arrays.h`. Subclasses will be lumped in later PRs with their parent classes to test their additional functionality.

#### `Tensor3D` tests | closes #156 

- Tests the default constructor
- Checks the `allocate` method can be used to initialise memory for the tensor
- Checks the `frobenius` on the zero-tensor is 0, and that `zero` sets all allocated entries to 0
- Tests a pre-built array (which presently is a MATLAB array loaded into the simulation) can be used to construct a `Tensor3D`
- Tests the `frobenius` function on a pre-built array that has been used to construct a `Tensor3D`
- Checks that the destructors can be called successfully.

#### `XYZTensor3D` tests | closes #152 

- Tests the default constructor
- Tests that the XYZ components can be accessed through the `AxialDirection` enum
- Tests that the XYZ components can be accessed by providing `'x','y','z'` `chars`
- Tests that, post-allocation, elements can be assigned and accessed

#### `XYZVectors` tests | closes #153 

- Tests the default constructor
- Tests `set_ptr` for both `AxialDirection` and `char` arguments
- Tests that arrays are correctly assigned to the `x`, `y`, `z` variables and these can be manipulated

#### `Matrix` tests | closes #154 

- Tests the default constructor
- Tests the overloaded constructor (where dimensions can be passed in)
- Checks that once allocated, the elements can be assigned to correctly

#### `Vector` tests | closes #155 

To be honest, I don't actually know why we have this class. `std::vector` is probably going to offer everything we want this to have, with the exception of the writing from MATLAB memory which we will need to add ourselves. (So I guess we could subclass std::vector?).

- Tests the default constructor
- Tests (to some extent) the MATLAB assignment method


tdms/tests/unit/array_tests/test_XYZTensor3D.cpp
```diff
@@ -0,0 +1,43 @@
...
+        for (int i = 0; i < n_rows; i++) {
+          // shouldn't seg fault
+          REQUIRE_NOTHROW(default_constructor_test[component][k][j][i] = 0.);
+        }
+      }
+    }
+  }
+  // at end, destructor should be called as appropriate
+  // but this class has no destructor --- it doesn't store the size of its arrays
+  // so we should use Tensor3Ds for the components no??
```
samcunliffe, line 42: Right....

> ### XYZTensor3D tests
>
>   * Tests the default constructor
 >  * Tests that the XYZ components can be accessed through the AxialDirection enum
>   * Tests that the XYZ components can be accessed by providing 'x','y','z' chars
>   * Tests that, post-allocation, elements can be assigned and accessed

I now can't find where these guys are explicitly `free`d? If the destructor doesn't exit it will be the default, and presumably won't deallocate all of the stuff we've allocated in `XYZTensor::allocate`. Do we explicitly free it all in `iterator.cpp`?

 `malloc` with no `free` ? Or do I miss something?

samcunliffe: If we're templating it... perhaps we want to try calling allocate on a `XYZTensor<std::complex>` and check for memory leaks.

tdms/tests/unit/array_tests/test_XYZVectors.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+  // now let's try assigning
+  v.set_ptr(AxialDirection::X, x_vec);
+  REQUIRE(v.x != nullptr);
+  v.set_ptr(AxialDirection::Y, y_vec);
+  REQUIRE(v.y != nullptr);
+  v.set_ptr(AxialDirection::Z, z_vec);
+  REQUIRE(v.z != nullptr);
+  // and manipulating the components
+  for (int ix = 0; ix < nx; ix++) { x_tot += v.x[ix]; }
```
samcunliffe, line 39:
```suggestion
  double x_tot = 0.;
  for (int ix = 0; ix < nx; ix++) { x_tot += v.x[ix]; }
```
For clarity, I might have declared the total(s) just here above the loop. Then delete space so l27 and lines 31-37 are all part of the same "block".

tdms/tests/unit/array_tests/test_Tensor3D.cpp
```diff
@@ -0,0 +1,60 @@
...
+ * @author William Graham (ccaegra@ucl.ac.uk)
+ * @brief Tests the functionality of the Tensor3D class, which is the building block for several further field classes
+ *
+ */
+#include <catch2/catch_test_macros.hpp>
+#include <spdlog/spdlog.h>
+
+#include "arrays.h"
+
+const double tol = 1e-8;
```
samcunliffe, line 12: Why not smaller? We're assigning and summing doubles, right?

samcunliffe (commented):
Not sure we should generalise/add templating at the same time as testing existing functionality. But I can see why you'd do it.

I'd like to see a testing of other types we might make arrays of. I mean: `std::complex` and maybe also `fftw_complex` (assuming that makes sense).

And you had to move some code?

tdms/tests/unit/array_tests/test_XYZVectors.cpp
samcunliffe: I wondered also, about using `std::accumulate` here. Might simplify a bit.

tdms/tests/unit/array_tests/test_XYZTensor3D.cpp
willGraham01: > `malloc` with no `free` ? Or do I miss something?

As a standalone class, this is a `malloc` with no free since there's no custom destructor. Also, for whatever reason we _don't_ store the dimensions of the XYZTensor3D arrays we allocate...

However, we only actually use `XYZTensor3D` as an attribute of `Field`, which _does_ free the memory we assigned in it's [custom destructor](https://github.com/UCL/TDMS/blob/1dfd74e37ff5c16818921d17393d36c1d79b927c/tdms/src/fields/base.cpp#L62-L70). Although that being said, there's no check in `Field` that this memory has been _assigned_ in the first place either :cry: 

I'll add a custom destructor to `XYZTensor3D` (_and_ check the other classes for potentially similar issues) that
1. Checks whether memory has been assigned
2. If so, free it on destruction
3. Otherwise, silently exit
`Field` will likewise receive an update.

tdms/src/arrays.cpp
```diff
@@ -212,49 +193,6 @@ Pupil::~Pupil() {
...
-double Tensor3D<complex<double>>::frobenius() {
-  double norm_val = 0.;
-  for (int i1 = 0; i1 < n_layers; i1++) {
-    for (int i2 = 0; i2 < n_cols; i2++) {
-      for (int i3 = 0; i3 < n_rows; i3++) { norm_val += norm(tensor[i1][i2][i3]); }
-    }
-  }
-  return sqrt(norm_val);
-}
-
```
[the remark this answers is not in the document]
willGraham01: Now it's templated, it belongs in the header. If you mean the `sqrt(norm_val)` part for `std:complex` inputs specifically, `std::abs` returns what we want without needing to take the square-root explicitly  and hence the whole thing can be templated.

tdms/tests/unit/array_tests/test_XYZTensor3D.cpp
```diff
@@ -0,0 +1,59 @@
...
+using namespace std;
+const int REAL = 0, IMAG = 1;
+
+TEST_CASE("Test XYZTensor allocation") {
+  SPDLOG_INFO("== Testing XYZTensor3D allocation/deallocation");
+  // dimensions for the test-tensor
+  int n_layers = 4, n_cols = 8, n_rows = 16;
+
+  // try creating an XYZTensor3D using the default constructor
+  XYZTensor3D<complex<double>> default_constructor_test;
```
samcunliffe, line 22: Marginal comment: worth to _also_ test double? Any value in testing both?
If easy in catch2, I'd say do it. If not: not.

samcunliffe (approved):

PR #149, opened 2022-10-27
Adding a `CellCoordinate` class to group `i,j,k`

### LINKED ISSUE #132: Remove `int i, int j, int k` data-clump

willGraham01:
Closes #132 | Depends on #151  | 

## `CellCoordinate` class

The `CellCoordinate` class is added, which allows us to avoid passing the `i,j,k` data clump to the various interpolation schemes called in `iterator.cpp`. It also clears up the function declarations in `Field`, `SplitField` and their subclasses.

For readability purposes, within the various interpolation functions it is clearer to unpack the cell indices rather than call `cell.{i,j,k}()` every time we want to access one of the members. To this end, there is an `unpack` method for this class which allows for easy export of the cell indices to local variables.

### Testing

This class is very simple, being more for readability aid than any actual functionality itself. The choice to use a class rather than a `struct` is because we might want to add functionality later: EG to provide the physical coordinates of the field components of a particular cell, which is currently managed _somewhere_ in `iterator.cpp`.

This being said, this class sees use in the `test_field_interpolation.cpp` test (where interpolation calls are needed).


willGraham01:
Making the diff more manageable by targeting the dependant branch.

tdms/include/cell_coordinate.h
```diff
@@ -25,4 +37,15 @@ class CellCoordinate {
...
+  /**
+   * @brief Copies the values of the cell indices into the addresses provided
+   *
+   * @param[out] i,j,k Address to place the value of cell_{i,j,k} into (respectively)
+   */
+  void unpack(int *i, int *j, int *k) {
+    *i = cell_i;
+    *j = cell_j;
+    *k = cell_k;
+  }
```
samcunliffe, line 50: I'd be a bit wary of [speculative generality](https://refactoring.guru/smells/speculative-generality) 👻 
What method(s) are we _actually_ going to want in the future that isn't just simple public-members `struct` behaviour?

samcunliffe: You can even add a constructor for the `struct` if you don't like `MyStruct flibble = {foo, bar};` ...
```c++
#include <iostream>

struct CellCoordinate {
  int i, j, k;
  CellCoordinate(int a, int b, int c): i(a), j(b), k(c) {}
};

std::ostream& operator<<(std::ostream& os, const CellCoordinate& c) {
    os << c.i << ',' << c.j << ',' << c.k;
    return os;
}

int main() {
    // standard way:
    //CellCoordinate c = {1,2,3};
    // also works:
    CellCoordinate c(4,5,6);
    std::cout << c << "\n";

    int i=c.i,j=c.j,k=c.k;
    std::cout << i << ", " << j << ", " << k;
    return 0;
}
```

tdms/src/fields/magnetic.cpp
```diff
@@ -65,13 +66,15 @@ Now with approximations of Ha at each (a_i, cell_b + Db, c_k), we can pass this
...
-complex<double> MagneticField::interpolate_to_centre_of(AxialDirection d, int i, int j, int k) {
+complex<double> MagneticField::interpolate_to_centre_of(AxialDirection d, CellCoordinate cell) {
   // this data will be passed to the second interpolation scheme
   complex<double> data_for_second_scheme[8];
   // this data will hold values for the interpolation in the first interpolation scheme
   complex<double> data_for_first_scheme[8];
   // the interpolation schemes that are to be used
   const InterpolationScheme *b_scheme, *c_scheme;
+  int i, j, k;
+  cell.unpack(&i, &j, &k);
```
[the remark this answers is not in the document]
samcunliffe: I'm just saying: your two lines w/ passing by reference vs. my one-liner.

samcunliffe:
Oh, I couldn't spot why the tests are failing. But it looks genuine? 🐛 🐞

tdms/include/arrays.h
```diff
@@ -223,7 +224,7 @@ class Tensor3D{
 
   void initialise(T*** tensor, int n_layers, int n_cols, int n_rows);
 
-  inline T** operator[] (int value) const { return tensor[value]; };
+  inline T **operator[](int value) const { return tensor[value]; };
```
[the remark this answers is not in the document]
willGraham01: VSCode seems to want to do this when I point Intelisense to `.clang-format`, so I assume so?

willGraham01:
> Oh, I couldn't spot why the tests are failing. But it looks genuine? bug lady_beetle

That's because the base branch, which is #149 is also failing due to using the new interpolation functions (same errors as #149's auto-checks). Since I need those new interpolation functions to be in place _before_ I remove the dataclump, these tests also fail.

I opened the PR against #149 so the diff was actually informative. Plan is to get Peter's approval on #149, merge that into main, wait for the reference outputs to be updated WRT the new interpolation scheme, then put this in.

tdms/include/cell_coordinate.h
willGraham01: ^ It looks pretty, ngl...

I feel like `cell_{i,j,k}` are useful as `private` attributes, since we don't plan to edit them and this stops us from doing that accidentally when looping/incrementing/interpolating over Yee cells. Maybe I'm being overly cautious though.

I also had the impression that somewhere in `iterator.cpp`, we need to acquire the spatial coordinates of the field values associated to each Yee cell, in which case we'd have a method on this class for doing that. But this might have already been eaten up in the refactor/ a figment of my imagination.

samcunliffe (approved):

willGraham01:
Per the plan, merging this into #151 since we're happy with the functionality and test failures are only due to test-reference mismatches that will need to be replaced with the underlying new interpolation functions.

PR #151, opened 2022-10-28
Replace cubic interpolation with Band-Limited Interpolation

### LINKED ISSUE #60: Odd behaviour in `interpolate.cpp`
# Odd behaviour in `interpolate.cpp` [Related to #44] 

Several functions in `interpolate.cpp` seem to have redundant `for` loops, and mismatches between the function doc-string and what the function actually does.
Looking at [`interpolateFieldCentralE_TE`](https://github.com/UCL/TDMS/blob/526987ce4ba47d982e90787b6e4e67af94dfb353/tdms/src/interpolate.cpp#L115-L176) for example:
- An error is thrown unless `k_u` $\leq 0\leq$ `k_l`, which is in contrast to what the docstring claims the acceptable values of `k_l` ( $\geq 2$ ) and `k_u` ( $\leq$ `K` $-2$) are.
- The variables `i_u`, `j_u`, `k_u` have intent[out], but are never assigned to, not pointers, and not returned.
- `for(k=k_l; k<=k_u; k++)` only executes when `k_u` $\leq 0\leq$ `k_l` due to the previous error checks, and so `k` can only take the value 0.

The same things occur in `interpolateFieldCentralE_TM`, `interpolateFieldCentralH_TE`, and `interpolateFieldCentralH_TM`.

If the intended behaviour is what's listed in the doc-strings; the code needs to be changed and we need to check whether the result of any demonstration code or tests _depend_ on then also changes, due to dependence on these functions.
If what's currently in the code is the intended behaviour, it looks like it can be made more readable and the doc-strings updated to match.

### LINKED ISSUE #113: Change the interpolation calls in `exphasorssurface`, `exphasorsvolume`, and `fieldsample` to the new interpolation functions.

### LINKED ISSUE #115: Remove now-redundant functions (and headers).

willGraham01:
Closes #113 | Closes #115 - at least in the interpolation context, which was the original context of the issue, as #44 tracks it | Closes #60 - since we've now overhauled how the interpolation works anyhow.

### Replaces Interpolation Functions

The old cubic interpolation methods are replaced with their band-limited interpolation equivalents, which are now function methods for the various `Field` and `SplitField` objects.

#### Impacts on Test Performance and Results

System tests are expected to fail due to the reference-runs they were generated from using the old cubic interpolation schemes. A sample visual comparison of the outputs (`arc_01-cyl`) is in the image below.
old-vs-new-arc_01-cyl

Below is a sample from various system tests of the "most different" (in terms of greatest mean-square-difference) outputs from their corresponding reference. These MSDs are of the order $10^{-1}$ in the worst case. The general shape of the computed fields remains the same, with slight numerical discrepancies.
MSD-Worst-Compare1

There are also MATLAB files that can produce these plots for any output file that is produced - these can be added to the repository if desired (although these are results-parsing, rather than central to the TDMS executable's functionality).

#### Unused files and cleanups (#115)

The `InterpolationScheme` class and `best_scheme` function supersede the original cubic-interpolation functions. 

- `interpolate.h` and `interpolate.cpp` are no longer needed and have been deleted. 
- `CMakeLists.txt` has been updated accordingly.
- The old cubic methods `interp{1,2,3}` and `CheckInterpolationPoints` have been deleted, as these are no longer required.
- `test_interpolation_functions.cpp` no longer tests the old `interp{1,2,3}` methods.
- `test_interpolation_determination.cpp` no longer runs `CheckInterpolationPoints`, since this has been deleted as it is unused.


willGraham01:
@prmunro Those two are `arc_03` - the first being the `sph` test and the latter the `fs` test. Below are all of the field components for those two tests (and not just the "worst" one):

`fs` test output:
arc_03-fs-outputs

`sph` test output:
arc_03_sph-outputs

Looks like Band-limited interpolation is "smoothing" out those larger amplitude spikes?

samcunliffe (approved):
Already approved as #143.
Still to resolve:

- [ ] Differences in plots.
- [ ] ...what tolerance is acceptable.

willGraham01:
> ...what tolerance is acceptable for the relative mean squared difference?

Once we're happy that the BLi functions are working, we'll need to replace the old reference files. From there on out, we can continue using `1e-10`.

As for the situation we're in now (where we're comparing to files generated by different interpolation methods), no real idea what we should be looking for :confused: My best stab at an idea is to obtain a distribution/quantification of the average difference between the results of BLi and cubic interpolation. EG If we knew the pointwise difference between BLi and cubic interpolation was distributed as $\mathcal{N}(\mu,\sigma)$, then
$$MSD \approx N_{\text{interpolation points}}^{-1} \sum_{i=1}^{N_{\text{interpolation points}}}\mu^2 = \mu^2$$
for example.

prmunro:
> Already approved as #143. Still to resolve:
> 
> * [ ]  Differences in plots. [Comment](https://github.com/UCL/TDMS/pull/143#issuecomment-1294768436), [comment](https://github.com/UCL/TDMS/pull/143#issuecomment-1294781114).
> * [ ]  ...what tolerance is acceptable for the relative mean squared difference?
> 
> For the latter, 1e-10 → 1e-6 feels OK, and like we're just taking a different interpolation into account. 1e-2 is potentially a problem. But that's a very unscientific gut feeling.

Hi Sam and Will, I have generated a test case which will show whether the new interpolation scheme is working. These tests have a much higher value of mean square error since we're limited by the accuracy of the PSTD method, rather than than machine precision.

You can download the files using the dropbox link below. As usual, run the script run_pstd_bscan.m. There are comments in the script explaining it. Also, if you want to re-run parts of the script you might have to set a variable back to logical 1. It would be interesting to see the value of variable errnum when the PSTD output is calculated using the new version of interpolation. 

*** Details for retrieval: ***
    The recipient(s) will need the claim passcode '4d111c81' 
to retrieve the file(s).  

pick-up URL: https://wwwapps-live.ucl.ac.uk/cgi-bin/dropbox/dropbox-https.cgi?state=pickup_info&id=66b93236
claim id: 66b93236
file basename: arc_14.zip

willGraham01:
@prmunro @samcunliffe Inital results from the new (band-limited) interpolation against the old interpolation schemes.

Significant (as in, one order of magnitude) improvements to the error when using the new interpolation functions on the $E_z$, $H_x$, $H_y$ components as we expect. A slight improvement for the other components.

Although the error still isn't $o(10^{-3})$, or approximately 1%.

image

A promising indication - once we have `fieldsamples` and `camplitudes` reference data we'll hopefully see similar improvements.

willGraham01:
### Results from the new test data in `arc_14`

Main takeaway is that, on the whole, BLi seems to be performing better than cubic in the majority of cases, and when it doesn't the additional error is $O(10^{-4})$.

| Error metric | # pts BLi is better / Total elements in error array | Worst error increase |
| - | - | - |
| `err_an` | 3 / 3 (100.00 %) | -5.55454e-04 |
| `err_csample` | 34 / 48 (70.83 %) | 8.67484e-05 |
| `err_fsample` | 1 / 1 (100.00 %) | -3.08566e-12 |
| `err_num` | 6 / 6 (100.00 %) | -4.98942e-04 |
| `err_surf` | 70 / 96 (72.92 %) | 7.48017e-04 |

If the entry in the last column is negative means that BLi was always superior to cubic interpolation. Otherwise, that entry is the additional error that BLi incurred _on top of_ that from the cubic interpolation. IE in `err_csample`, BLi incurred errors no greater than 8.67484e-05 of those obtained from cubic interpolation.

The script to generate these figures (and the raw error outputs from running arc_14 with the old and new interpolation functions) is below.
[arc_14_results.zip](https://github.com/UCL/TDMS/files/10133921/arc_14_results.zip)

prmunro:
Hi @willGraham01 , this is a good result. Thanks for including the error data and script. It was good to see that in all cases, the new interpolation scheme brings the error for the Ez, Hx and Hy components down to more or less the same order of magnitude as the other components. It's curious to me as to what is going on away from the central wavelength (so columns with a low or high index), however, I don't think this is a result of the interpolation, rather something to do with the PSTD method. I'd like to look into that at some point. I have another test complete which I'll upload today. After looking at this results, I think we'll be in a position to approve the interpolation PR.

willGraham01:
Now also resolves #132

prmunro (approved):
Happy for this to be merged on the basis of substantial testing carried out.

samcunliffe:
Good work @willGraham01.

PR #182, opened 2022-11-13
Fix the MacOS build yet again.

samcunliffe:


.github/workflows/macos_tests.yml
```diff
@@ -50,6 +50,11 @@ jobs:
           brew unlink libomp
           brew link --force libomp
 
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.10"
```
samcunliffe, line 56: The problem is restricted to python/pip/h5py 3.11](https://github.com/samcunliffe/h5py-macos-ghrunner-mwe/actions/runs/3456736301).

samcunliffe:
Unrelated, but the bypass merge option is was live. 
I've turned it off in settings.

Screenshot 2022-11-13 at 20 57 56

.github/workflows/macos_tests.yml
willGraham01: We have #183 to remind us to check-in occasionally to see if the upstream problem has been fixed (and to remind us why we're effectively forcing a Python 3.10 install)

willGraham01 (approved):

PR #187, opened 2022-11-14
Expand `field.h` unit tests

willGraham01:
Moves unit tests for `field` classes into their own sub-folder, and expands on the tests we currently have.


tdms/include/field.h
```diff
@@ -322,7 +328,9 @@ class MagneticField: public Field{
  * Structure to hold a field and allow saving it to a file
  */
 class TDFieldExporter2D{
-
+private:
+  int nI = 0;
+  int nK = 0;
```
samcunliffe, line 333:
```suggestion
  int nI = 0; /**< documentme */
  int nK = 0; /**< documentme */
```

tdms/include/field.h
```diff
@@ -331,7 +339,7 @@ class TDFieldExporter2D{
   /**
    * Allocate the arrays to hold the field
    */
-  void allocate(int nI, int nJ);
+  void allocate(int _nI, int _nK);
```
samcunliffe, line 342: Fun fact. Google style prefers a **trailing** underscore for private members. So it should be the other way around: the public API gets the names unadorned, and the private class workings get underscores.

I don't propose you change here though. All of our classes break this rule. (Perhaps worth enforcing globally at some stage. Low priority for now though.)

tdms/include/unit_test_utils.h
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * @brief Determines if two numerical values are close by relative comparison.
+ *
+ * Checks the truth value of the condition
+ * abs(a - b) / max(abs(a), abs(b)) < tol.
+ * If a and b are both close to zero, returns true.
+ *
+ * @param a,b Numerical values
+ */
+inline bool is_close(T a, T b, double tol=1E-10, double close_to_zero_tol=1E-30) {
```
samcunliffe, line 17:
```suggestion
/**
 * @brief Determines if two numerical values are close by relative comparison.
 *
 * Checks the truth value of the condition
 * abs(a - b) / max(abs(a), abs(b)) < tol.
 * If a and b are both close to zero, returns true.
 *
 * @param a,b Numerical values
 */
template<typename T>
inline bool is_close(T a, T b, double tol=1E-10, double close_to_zero_tol=1E-30) {
```

Check for me that this works? But I thought doxygen was smart enough to understand templating.

tdms/include/unit_test_utils.h
```diff
@@ -0,0 +1,41 @@
...
+
+template<typename T>
+/**
+ * @brief Computes the Euclidean norm of the vector provided
+ *
+ * @param v Vector or array
+ * @param end (Inclusive) end of buffer to read vector from
+ * @param start (Inclusive) start of buffer to read vector from
+ * @return double Euclidean norm
+ */
```
samcunliffe, line 36:
```suggestion
/**
 * @brief Computes the Euclidean norm of the vector provided
 *
 * @param v Vector or array
 * @param end (Inclusive) end of buffer to read vector from
 * @param start (Inclusive) start of buffer to read vector from
 * @return double Euclidean norm
 */
template<typename T>
```

tdms/tests/unit/field_tests/test_Field.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+#include <catch2/catch_test_macros.hpp>
+
+#include "globals.h"
+#include "unit_test_utils.h"
+
+using namespace std;
+using namespace tdms_math_constants;
+
+const double tol = 1e-16;
```
samcunliffe, line 19:
```suggestion
const double TOLERANCE = 1e-16;
```

tdms/tests/unit/field_tests/test_Field.cpp
```diff
@@ -0,0 +1,112 @@
...
+                (x_normalised_correctly &&
+                 (abs(F.real.x[k][j][i] - 1. + IMAGINARY_UNIT * (F.imag.x[k][j][i] - 0.)) < tol));
+        // set y component to be 2, so should normalise to be 1 - i
+        y_normalised_correctly =
+                (y_normalised_correctly &&
+                 (abs(F.real.y[k][j][i] - 1. + IMAGINARY_UNIT * (F.imag.y[k][j][i] + 1.)) < tol));
+        // keep z components as 0
+        z_normalised_correctly =
+                (z_normalised_correctly &&
+                 (abs(F.real.z[k][j][i] - 0. + IMAGINARY_UNIT * (F.imag.z[k][j][i] - 0.)) < tol));
```
samcunliffe, line 65:
```suggestion
        x_normalised_correctly &= (abs(F.real.x[k][j][i] - 1. + IMAGINARY_UNIT * (F.imag.x[k][j][i] - 0.)) < TOLERANCE);
        // set y component to be 2, so should normalise to be 1 - i
        y_normalised_correctly &= (abs(F.real.y[k][j][i] - 1. + IMAGINARY_UNIT * (F.imag.y[k][j][i] + 1.)) < TOLERANCE);
        // keep z components as 0
        z_normalised_correctly &= (abs(F.real.z[k][j][i] - 0. + IMAGINARY_UNIT * (F.imag.z[k][j][i] - 0.)) < TOLERANCE);
```

samcunliffe: But also, you could do the `REQUIRE` for each iteration in the loop? Or is that slower?

tdms/include/unit_test_utils.h
```diff
@@ -0,0 +1,41 @@
+/**
+ * @file unit_test_utils.h
+ * @brief Functions common to multiple unit tests
+ */
```
samcunliffe, line 4: I was also wondering about pulling out some test utils into a (header, src) pair. My (very very soft) problem is that I'd prefer testing to live together.

Can we have the header in `tests/unit`? Is our build system smart enough for that?

tdms/src/fields/td_field_exporter_2d.cpp
```diff
@@ -1,21 +1,28 @@
...
+#include "matlabio.h"
 
-void TDFieldExporter2D::allocate(int nI, int nJ) {
+void TDFieldExporter2D::allocate(int _nI, int _nK) {
 
-  mwSize dims[2] = {nI, nJ};
+  nI = _nI;
+  nK = _nK;
+  mwSize dims[2] = {_nI, _nK};
   matlab_array = mxCreateNumericArray(2, (const mwSize *)dims, mxDOUBLE_CLASS, mxREAL);
```
samcunliffe, line 12:
```suggestion
  mwSize dimensions[2] = {_nI, _nK};
  matlab_array = mxCreateNumericArray(2, (const mwSize*) dimensions, mxDOUBLE_CLASS, mxREAL);
```

tdms/include/unit_test_utils.h
```diff
@@ -0,0 +1,41 @@
...
+#include <complex>
+
```
samcunliffe, line 6: P.s. we should also pull out the global TOLERANCE and define it once here in this file (probably need to find-replace).
```suggestion
#include <complex>

namespace tdms_tests {
const double TOLERANCE = 1e-16;
}
```

samcunliffe: Actually now I come to think of it, I'd also probably put all of these functions into the namespace.
So `tdms_tests::is_close` etc. Then it's clearly separated from any catch2 stuff.

tdms/tests/unit/field_tests/test_TDFieldExporter2D.cpp
```diff
@@ -0,0 +1,33 @@
...
+TEST_CASE("TDFieldExporter2D") {
+
+  // create a 0-field for us to export
+  const int I_tot = 8, J_tot = 1, K_tot = 8;
+  ElectricSplitField Es(I_tot, J_tot, K_tot);
+  Es.allocate_and_zero();
+
+  // create our field exporter
+  TDFieldExporter2D tdfe2d;
+  tdfe2d.folder_name = ".";
```
samcunliffe, line 18: We need write to the current working directory to run the tests?

samcunliffe, line 18:
```suggestion
  tdfe2d.folder_name = "."; // TODO: Sam has some temp directory code that should be used here. TDMS/#80.
```
This means we need write permission to the current working directory to run the tests?

samcunliffe (approved):

tdms/tests/unit/field_tests/test_Field.cpp
willGraham01: A *bit* slower. It was more so that we don't end up bloating how many checks we are doing in this test. 

My ethos was I'm testing whether each of `x`, `y`, `z` have had _all_ their components normalised correctly, so I should have 3 `REQUIRES`. Doing the `REQUIRES` in the loop just bloats the number of checks we do; I feel $2/3$ tests failing is more informative than `X / I_tot*J_tot*K_tot`.

tdms/tests/unit/field_tests/test_TDFieldExporter2D.cpp
willGraham01: Afraid so - couldn't think of a suitable place to put the output that is going to be produced. If you have some `temp` directory stuff, I agree we can hijack that later.

tdms/include/unit_test_utils.h
willGraham01: Let me experiment on this branch, after looking at the other comments :+1:

tdms/tests/unit/field_tests/test_TDFieldExporter2D.cpp
samcunliffe: When this is merged I'll cherry-pick it out of my feature branch and see if I can make a small self-contained PR.

tdms/include/unit_test_utils.h
willGraham01: Turns out `CMake` is smart enough [if you tell it where to look](https://github.com/UCL/TDMS/pull/187/files#diff-a8a22f01e97802a8ec881f48890a12b6feb81a367999801dbb0feb31ffbf47d5R144-R146).

Issue #181, opened 2022-11-11
Use HD5 input and library


willGraham01:
The project currently makes use of `.mat` files and `MATLAB` to open the input and output files.
With a view towards #70, we need to move away from this dependency.

One such option is `libhd5`, as `.mat` files are essentially already `hd5` files but under a different name. Changing our input interpreter will have several knock-on effects. A list (that is potentially non-exhaustive) of the key changes to implement is below:

- [ ] The classes in `arrays.h` and `field.h` need to be updated to read in data from `hd5` files
- [ ] The interpolation functions no longer need to assign to temporary MATLAB arrays, instead to the `HD5` outputs
- [ ] `execute_simulation` can potentially have its input arguments overhauled, now that we aren't limited to 4 inputs (*cough* `mex` *cough*)

samcunliffe:
Not completely replacing MATLAB yet. But close. 
See also
* #285

PR #189, opened 2022-11-17
A temporary directory utility function for file I/O tests.

samcunliffe:
Adding tmp directory function which was developed for #181, but seems like it's useful already.

@alessandrofelder mentioned that there might be a cleaner(?) way to do this in CMake (which I haven't investigated). But mine is now only a few lines of C++ from the `std` library. So hopefully not too clunky.

***

The `*_utils.h` file has a messy diff because I changed the indentation. I'd recommend [turning whitespace off](https://github.blog/2018-05-01-ignore-white-space-in-code-review/) (if you haven't already done so).

### DIFF
--- a/tdms/tests/include/unit_test_utils.h
+++ b/tdms/tests/include/unit_test_utils.h
@@ -5,12 +5,15 @@
 #pragma once
 
 #include <complex>
+#include <filesystem>
+#include <random>
+#include <string>
 
 namespace tdms_tests {
 
-  inline double TOLERANCE = 1e-16; //< Floating-point comparison tolerance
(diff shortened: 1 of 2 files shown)


tdms/tests/include/unit_test_utils.h
```diff
@@ -5,12 +5,15 @@
 #pragma once
 
 #include <complex>
+#include <filesystem>
+#include <random>
+#include <string>
 
 namespace tdms_tests {
```
samcunliffe, line 12: Namespaces shouldn't be indented according to [Google](https://google.github.io/styleguide/cppguide.html#Namespaces).

willGraham01 (approved):

PR #192, opened 2022-11-22
Samcunliffe array test class

willGraham01:
Fixes the MacOS requirement that `const static` be used in place of `const` for private class attributes that are used to declare arrays.


samcunliffe:
\o/

samcunliffe (approved):

PR #193, opened 2022-11-29
181 hdf5 io

samcunliffe:
[Draft] of adding hdf5. Tests should fail because: TDD. No functional code in these commits.


samcunliffe:
Closing this PR, but keeping the branch around.
This was just a demo/PoC so can equally well be archived in the page of declined PRs and it tidies up the open list.

giordano:
> but keeping the branch around.

Deleted branches can always be restored at any time if really needed.

Issue #195, opened 2022-11-30
Interpolation method as a toggle


willGraham01:
## 29/11/2022

It was asked whether it would be possible to let the user decide whether to use the Band-Limited interpolation methods, or whether to revert to using the old cubic interpolation, through the input file.

We can break this down into two stages:
- [ ] (Pre-#70) Make this an option that is configurable from the command-line
- [ ] (Post-#70) Make this flag something which is read in from the input file

PR #196, opened 2022-11-30
Interpolation method toggle

willGraham01:
Resolves the first objective of #195.

Updates a couple of the tests which check that the interpolation schemes are being correctly selected, to check that the over-write flag works.

TODO: Make the CI workflows have matrix options to check the cubic and BLi reference outputs.


samcunliffe:
Strange problem with the diff between these two branches.
Closing here to see if reopening a PR between them fixes this.

PR #197, opened 2022-12-05
Reduce code duplication in unit tests introduced by #80

### LINKED ISSUE #125: Reduce code duplication in unit tests
The unit tests introduced to test the functionality of the interpolation schemes (#80) suffer from a lot of code repetition and function name repetition.

We should:
- [ ] Break the commonly-used `inline` functions out into their own supporting file. These functions could go into their own header/`.cpp` file, or into the `interpolation_methods` files (since they will be used for _testing_ said methods.
- [ ] Reduce the number of loops and general code repetition in `test_field_interpolation.cpp` and `test_interpolation_functions`. The loops in question _are not completely identical_, but are similar enough that condensing them should be possible.

willGraham01:
Uses `Catch2`'s sectioning abilities to reduce the amount of code repetition in the tests that were introduced in #80. Closes #125 .

This is mainly just using sections to breakout the few portions of each of these tests that are different - creation of the data, the exact function values, etc.

Sadly it's impossible to go the full throw and drop the distinction between complex-valued and real-valued function tests (unless we want to treat "real" functions as "complex" functions with no imaginary part, but then we aren't testing the time-domain stuff).

**Half the diff** is an indentation bug in `unit_test_utils.m`. Not sure why this appears.

### DIFF
--- a/tdms/tests/include/unit_test_utils.h
+++ b/tdms/tests/include/unit_test_utils.h
@@ -11,9 +11,17 @@
 
 namespace tdms_tests {
 
-inline double TOLERANCE = 1e-16;//< Floating-point comparison tolerance
+  inline double TOLERANCE = 1e-16;//< Floating-point comparison tolerance
 
-/**
+  /**
+  * @brief Determines whether a value is close to zero
+  *
+  * @param x Value to test
+  * @param tol Tolerance for being "close" to zero
+  */
+  inline bool near_zero(const double &x, double tol=TOLERANCE) { return std::abs(x) < tol; }
+
+  /**
  * @brief Determines if two numerical values are close by relative comparison.
  *
  * Checks the truth value of the condition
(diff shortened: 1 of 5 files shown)


samcunliffe (approved):
All of this looks good. Why still a draft? To my taste it could be merged.

> Sadly it's impossible to go the full throw and drop the distinction between complex-valued and real-valued function tests (unless we want to treat "real" functions as "complex" functions with no imaginary part, but then we aren't testing the time-domain stuff).

Yeah. Gottit. I'd propose we leave it as-is for now. We can think about being smarter later.

The indentation is a difference between your and my editor setup, or maybe our clang settings. FWIW, according to Google: namespaces shouldn't be indented. 

(But this is one of those things to fix later with a global style unification as part of #41.) 

The diff can be de-cluttered by clicking on the ⚙️ > little popup to choose how you want to see the diff > option at the bottom "Hide whitespace".

samcunliffe (approved):

Issue #116, opened 2022-10-04
Choose a consistent file structure for integration test data


dstansby:
In https://github.com/UCL/TDMS/pull/101 I'm trying to simplify the integration tests so it's really easy to add new tests. I've noticed that there is not a consistent file structure in the test files that have been uploaded. For the low numbered tests (01, 02, 03) the structure is

- a single input file named `pstd_input_file.m`
- one or more input data files named `pstd_{run_type}_input.mat`.
  run_type can be any string.
- corresponding reference output data files named `pstd_{run_type}_reference_output.mat`.
  For every input file there is a corresponding output file with the same run_type in the filename.

But in the high nubmered tests (09, 10) there are "in" and "out" directories. To make the test infastructure as simple as possible, it would be good to choose, document, and stick to a file structure for the integration test files.

samcunliffe:
Do you have a preference? 

Peter uses directories `in/` and `out/` in his tarballs so
* [Pro] it's in keeping with his workflow & scripts.
* [Con] `*_reference_output.mat` is less explicit.

dstansby:
I don't personally have a preference and think either is fine, so I'd say go with what Peter prefers since he'll be using the software after the ARC project finishes

samcunliffe:
Solved by @willGraham01 (quite a while ago now... I've lost the PR)

willGraham01:
This feature comes in with #236, however reaches it's final (and present solution state) in #244.

The `config_XX.yaml` files explicitly provide the full path to necessary input/output/reference files for each test.

PR #202, opened 2022-12-14
Update system tests

willGraham01:
# Updates the system tests to include the cubic/band-limited toggle

Every system test that we were previously running now has a counterpart:
- One run using the cubic interpolation
- One run using the band-limited interpolation

Previously, the `tdms` executable used cubic interpolation, with the exception of using band-limited interpolation for one particular portion of extracting the `fieldsample`. This was a fixed workflow.

Now that `tdms` takes a toggle to switch between cubic and band-limited interpolation, the reference test data needs to be re-generated. This has been done and new Zenodo links are pointed to.

## Test parametrisation

Along similar lines to #101, some effort has been made at parametrising the new system tests to account for the fact that each `.zip` folder contains between 2 and 4 tests. Each `.zip` file now contains a `config.yaml` file at it's top-level, which has the syntax:

```yaml
test_id: XX
tests:
    test_name_without_spaces_1:
        input_file: relative_path_to_input1
        reference: relative_path_to_reference_output2
        fdtd_solver: True/False [optional, default is False is absent]
        cubic_interpolation: True/False [optional, default is False if absent]
    test_name_without_spaces_2:
        input_file: relative_path_to_input2
        reference: relative_path_to_reference_output2
        fdtd_solver: True/False [optional, default is False is absent]
        cubic_interpolation: True/False [optional, default is False if absent]
```

The file `test_system.py` replaces the previous `test_arcXX.py` files. The `TEST_URLS` variable in `test_system.py` can instead simply be updated with an additional Zenodo url and test identifier and `pytest` parametrize magic will take care of the rest. There's also a `help()` function that goes into the details of writing the config files, so we have documentation somewhere.

As a result, this PR superceeds #101.


tdms/tests/system/read_config.py
```diff
@@ -0,0 +1,152 @@
...
+
+def validate_run_information(runs: dict) -> None:
+    for test_id, test_info in runs.items():
+        for required_info in run_required_info:
+            if required_info not in test_info.keys():
+                raise RuntimeError(f"{required_info} not provided in {test_id}")
+    return
+
+
+class TDMSSystemTest:
```
samcunliffe, line 38: Can haz docstring plz?

samcunliffe: _Might_ have made a bit of a boilerplate code saving by making this a [dataclass](https://docs.python.org/3/library/dataclasses.html) and writing a `__post_init__` to do the validation and throwing runtime errors.

Probably not worth changing now.

samcunliffe (approved):
I'd put a docstring in every function and class member. Then you may find you need fewer comments.

Functionally this looks ok though. You've deboilerplated a lot of the test boilerplate. 👍

tdms/tests/system/read_config.py
willGraham01: I have a local commit that adds docstrings to everything... just don't want to push it yet since since that'll make the tests restart and I want them to finish so they cache the data files :sweat_smile:

tdms/tests/system/read_config.py
```diff
@@ -0,0 +1,152 @@
...
+            input_file: pstd_fs_input.mat
+            reference: pstd_fs_cubic_reference.mat
+            cubic_interpolation: True
+        cyl_bli:
+            input_file: pstd_cyl_input.mat
+            reference: pstd_cyl_bli_reference.mat
+        cycl_cubic:
+            input_file: pstd_cyl_input.mat
+            reference: pstd_cyl_bli_reference.mat
+            cubic_interpolation: True
```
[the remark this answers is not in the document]
willGraham01: That's the plan/idea. In hindsight I could have just restructured the system data when I was regenerating it... but didn't want to hard-code in that kind of convention (although if you look at the `help()` function I stuck to a particular structure for the reference outputs at least).

At least this way, tests can be made in whatever structure you like and so long as you setup the config file correctly, pytest won't complain. #116 isn't entirely obsolete if we want to standardise the `.zip` file formats, but right now we have something that can handle these (somewhat trivial) file-structure differences.

PR #203, opened 2022-12-14
Introduce a versioning system.

### LINKED ISSUE #201: Versioning
As I'm writing doc for the CLI, it strikes me that we will want a) to tag versions of TDMS and b) a command-line `tdms --version`.

At the very simplest. we could just have a global `#define TDMS_VERSION` in a header all on its own, which is manually incremented following a `git tag`. But perhaps there's a better way.

TODO: investigate automated C++ versioning tools.

samcunliffe:
* Add a mechanism for CMake to read a git hash and construct a version. 
* A tag of semver format on main (v1.2.3) is taken as an **_official_** version, otherwise it constructs a placeholder <branchname>_<tag or short format commit hash>.
* (Useful for debugging) you can also overwrite this with CMake configuration...
   ```
   cmake .. -DTDMS_VERSION=Ive_just_introduced_a_bug
   make install
   tdms --version
   ```

With a bit of luck I've explained all of this in the developers documentation 😸 

***

Resolves #201.

***

Thanks to @giordano and @alessandrofelder for their suggestions.

### DIFF
--- a/doc/developers.md
+++ b/doc/developers.md
@@ -22,6 +22,34 @@ We depend on ...
 * [spdlog](https://github.com/gabime/spdlog) for logging (this is installed automatically by a CMake [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) if not found).
 * [MATLAB](https://www.mathworks.com/products/matlab.html) since we read in and write out MATLAB format files. We are actually thinking of removing this as a strict dependency ([#70](https://github.com/UCL/TDMS/issues/70)).
 
+### Versions
+
+We follow [semantic versioning](https://semver.org): major, minor, patch.
+Major versions may contain changes to the API and UI, minor versions contain new features but preserve backwards-compatibility of UI, patches are bugfixes.
+
+So (for example) version v1.2.0 is major version 1, minor version 2, and the zeroth patch. v1.2.0 will be backwards-compatible with v1.1.99.
+
+**Mechanics**: Version control is via `git tag`s on the `main` branch and propagated to the C++ source [code via CMake](https://github.com/UCL/TDMS/blob/main/tdms/cmake/version_from_git.cmake). Release notes are kept in the [GitHub release page](https://github.com/UCL/TDMS/releases).
+
+As well as official releases, a placeholder "version" is assigned to every commit on every development branch, or can be supplied at compile-time (which takes the highest precedence).
+
+```{.sh}
+$ cmake .. <usual cmake options> -DTDMS_VERSION=this_version_string_will_be_highest_priority
+$ make install
+$ tdms --version
+TDMS version: this_version_string_will_be_highest_priority
+```
+
+If on a development branch then the placeholder version is `<branchname>_<tag>` if tagged, or `<branchname>_<commit hash>` if not tagged. This is useful when debugging problems introduced between commits (with the assistance of `git bisect`).
+
+```{.sh}
+$ git checkout myfeaturebranch
+$ cmake ..
+$ make install
+$ tdms --version
+TDMS version: myfeaturebranch_a341ab
+```
+
 ### Code style and other admin
 
 We try to stick to [Google style C++](https://google.github.io/styleguide/cppguide.html) wherever practicable. <!-- And we run `clang-format` and `cppclean` linters as part of our CI. -->
--- a/tdms/CMakeLists.txt
+++ b/tdms/CMakeLists.txt
@@ -81,6 +81,17 @@ else()
     add_definitions(-DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_INFO)
 endif()
 
+
+# TDMS version ----------------------------------------------------------------
+# if supplied via the cmake configuration (-DTDMS_VERSION=whatever_I_want) then
+# respect that, otherwise construct a version from a git tag or commit hash.
+if (NOT TDMS_VERSION)
+    include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/version_from_git.cmake)
+endif()
+add_compile_definitions(TDMS_VERSION="${TDMS_VERSION}")
+message(STATUS "Compiling TDMS version: ${TDMS_VERSION}")
+
+
 # TDMS target -----------------------------------------------------------------
 
 # wildcard to find all *.cpp files in the src directory except for main.cpp
@@ -100,6 +111,8 @@ endif()
 
 # Compile options -------------------------------------------------------------
 target_compile_options(tdms PUBLIC -DMX_COMPAT_32 -c ${DFLAG} -O3)
+add_compile_definitions(TDMS_VERSION="${TDMS_VERSION}")
+#target_compile_definitions(tdms PRIVATE TDMS_VERSION="${TDMS_VERSION}")
 
 if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     target_compile_options(tdms PRIVATE /W4)
--- a/tdms/cmake/version_from_git.cmake
+++ b/tdms/cmake/version_from_git.cmake
@@ -1,0 +1,30 @@
+find_package(Git)
+if(GIT_FOUND)
+
+	# Get current branch name.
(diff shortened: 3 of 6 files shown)


samcunliffe:
It's a bit fiddly, but @willGraham01 : you can check out your local copy of `main` and merge this feature branch in (or cherry-pick the commit) to test.

Then `git tag v100.99.98` (or whatever) and compile as normal then `tdms --version`.

tdms/cmake/version_from_git.cmake
```diff
@@ -0,0 +1,30 @@
+find_package(Git)
+if(GIT_FOUND)
+
+	# Get current branch name.
```
giordano, line 4: What happens when one uses a snapshot, an archive of the source code without the git repository, e.g. https://github.com/UCL/TDMS/archive/refs/heads/201-versioning-in-cmake.tar.gz?

giordano (commented):
Looks overall good, but I believe you should handle the case where there is no git repository, which if I understand correctly (but haven't tried) may result in some errors when running CMake

tdms/CMakeLists.txt
```diff
@@ -81,6 +81,17 @@ else()
     add_definitions(-DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_INFO)
 endif()
 
+
+# TDMS version ----------------------------------------------------------------
+# if supplied via the cmake configuration (-DTDMS_VERSION=whatever_I_want) then
+# respect that, otherwise construct a version from a git tag or commit hash.
+if (NOT TDMS_VERSION)
```
willGraham01, line 88: One slightly annoying thing is that `-DTDMS_version` will be persistent between builds, and always take priority. 

This isn't a super big problem (since I presume we're always using fresh builds to test new features) but it does mean that if you provide a manual version number _once_ then you have to clear that variable from your `CMakeCache` manually. Is there a way to tell `cmake` to clear the manually stored variable / not place it in the cache?

willGraham01 (commented):
> It's a bit fiddly, but @willGraham01 : you can check out your local copy of `main` and merge this feature branch in (or cherry-pick the commit) to test.
> 
> Then `git tag v100.99.98` (or whatever) and compile as normal, then check `tdms --version`.

Seems all good: `cmake` happily picks up if I'm sat at a `tag`, and failing that provides the auto-generated version string you specified. `-DTDMS_version` overrides both as predicted, although I made a comment about how adamant it is about doing this between builds.

tdms/CMakeLists.txt
samcunliffe: Not without breaking the build. It needs to be in the cache (otherwise it won't get propagated at compile).

By-the-way this "problem" will also affect the release names in general... because it always needs to be cached. So you can break my system if you make a new commit and compile directly without rerunning the CMake configuration step, the "release" will be `<branchname>_<old commit hash>` for exactly the same reason.

tdms/cmake/version_from_git.cmake
samcunliffe: Ha. Good spot. I thought it would call it `"development"` from the `#ifdef` in the C++. But it calls it `_` (underscore).

Perhaps more important, is how to fix the git repo-less tarballs one gets from the [releases page](https://github.com/UCL/TDMS/releases). Presumably, we really want **those** to be correctly versioned.

tdms/CMakeLists.txt
willGraham01: No problem, we'll just have to remember to do clean builds when developing if we care about the version name. End-users should never encounter the problem if they're building from main and following our setup instructions anyway.

tdms/cmake/version_from_git.cmake
giordano: I don't think that one can possibly work unless the version information is stored in a file checked in the repository, or unless you automatically create with cmake a `VERSION` file which is not checked in the repository but would be included a in "release tarball" that you upload in the release page on GitHub, but that wouldn't be the automatic source tarball generated by GitHub for the tag.

The alternative to use automatic git tagging is to write down the version number in the CMake file, and maybe use to add the git hash only when you're building straight out of a git repository.

giordano: >  It obviously fails if the user untars to some other directory else.

Yeah, I don't like relying on the name of the parent directory, strictly speaking that's "outside" of the package (just slightly, it's on its edge).

> But I don't like hard-coding.

Well, if you want to make this work without a git repository this ***has*** to be hardcoded somewhere :slightly_smiling_face:

tdms/src/argument_parser.cpp
```diff
@@ -18,6 +19,11 @@ ArgumentNamespace ArgumentParser::parse_args(int n_args, char *arg_ptrs[]) {
     exit(0);
   }
 
+  if (args.have_flag("-v") || args.have_flag("--version")) {
+    print_version();
+    exit(0);
+  }
```
samcunliffe, line 25: Should unit-test this.

 - [ ] unit-test this.

doc/developers.md
```diff
@@ -19,7 +19,35 @@ We depend on ...
...
 
-## Code style and other admin {#code-style-and-doxygen}
+### Versions
+
+We follow [semantic versioning](https://semver.org): major, minor, patch.
+Major versions may contain changes to the API and UI, minor versions contain new features but preserve backwards-compatibility of UI, patches are bugfixes.
+
+So (for example) version v1.2.0 is major version 1, minor version 2, and the zeroth patch. v1.2.0 will be backwards-compatible with v1.1.99.
+
+**Mechanics**: Version control is via `git tag`s on the `main` branch and propagated to the C++ source [code via CMake](https://github.com/UCL/TDMS/blob/main/tdms/cmake/version_from_git.cmake). Release notes are kept in the [GitHub release page](https://github.com/UCL/TDMS/releases).
```
willGraham01, line 29:
```suggestion
**Mechanics**: Version control is via `git tag`s on the `main` branch and propagated to the C++ source [code via CMake](https://github.com/UCL/TDMS/blob/main/tdms/cmake/version_from_git.cmake).
Release notes are kept in the [GitHub release page](https://github.com/UCL/TDMS/releases).
```

tdms/include/argument_parser.h
```diff
@@ -131,6 +131,8 @@ class ArgumentParser {
 private:
   /** Prints the help message (all options).  */
   static void print_help_message();
+  /** Prints this software version. */
```
willGraham01, line 134:
```suggestion
  /** Prints the version of the compiled executable. */
```

There is no subject for "this" to refer to.

tdms/tests/unit/test_argument_parser.cpp
```diff
@@ -19,10 +19,18 @@ using std::vector;
...
+
+  SECTION("Help") {
+    const char *input_args[] = {"tdms", "-h"};
+    auto args = ArgumentNamespace(2, const_cast<char **>(input_args));
+    REQUIRE(args.have_flag("-h"));
+  }
+  SECTION("Version") {
+    const char *input_args[] = {"tdms", "--version"};
+    auto args = ArgumentNamespace(2, const_cast<char **>(input_args));
+    REQUIRE(args.have_flag("--version"));
```
willGraham01, line 32:
```suggestion
  SECTION("Help") {
    const char *input_args[] = {"tdms", "-h"};
    auto args = ArgumentNamespace(2, const_cast<char **>(input_args));
    REQUIRE(args.have_flag("-h"));
    REQUIRE(!args.have_flag("--help"));
  }
  SECTION("Version") {
    const char *input_args[] = {"tdms", "--version"};
    auto args = ArgumentNamespace(2, const_cast<char **>(input_args));
    REQUIRE(args.have_flag("--version"));
    REQUIRE(!args.have_flag("-v"));
```

B/c `ArgumentParser` isn't setting some member variable, so we should make sure that `have_flag` is being rigid.

willGraham01 (approved):

tdms/cmake/version_from_git.cmake
samcunliffe: Just to close the loop here (ancient post).
We'll hard-code the value as a fall back for now.

PR #204, opened 2022-12-16
Iterator refactor [3/N]

willGraham01:
Attempting to do some small refactors to `iterator.cpp` that don't touch anything that will need to be touched by #70. Docstrings are added to all refactored functions, classes, etc.

### `checkPhasorConvergence`

Replaced with the `Field` class method `max_pointwise_difference_over_max_element`. The name of the old function is misleading, it is only computing an error metric and NOT checking convergence - this happens later in the `iterator.cpp` main loop.

Pulling this out into a class method to make things a bit cleaner. Also, both the inputs are `Field`s, and the function is asymmetric, so it makes sense to have this as a class method.

### `setGridLabels`

Changed to the `initialise_from` method of the `GridLabels` class.

### `is_dispersive_ml`

Changed to `is_dispersive` method of `DispersiveMultilayer` class.

### `is_conductive`

Superceeded by the `all_elements_less_than` method of the `XYZVector` class. The check for whether a material is conductive is just whether at least one entry of an XYZVector exceeds a particular threshold.

### DIFF
--- a/tdms/include/arrays.h
+++ b/tdms/include/arrays.h
@@ -172,6 +172,16 @@ public:
 
 public:
   explicit DispersiveMultiLayer(const mxArray *ptr);
+
+  /**
+   * @brief Determines whether the (background) medium is dispersive
+   *
+   * @param K_tot Number of Yee cells in the z-direction (number of entries in this->gamma)
+   * @param near_zero_tolerance Tolerance for non-zero gamma (attenuation) values
+   * @return true Background is dispersive
+   * @return false Background is not dispersive
+   */
+  bool is_dispersive(int K_tot, double near_zero_tolerance=1e-15);
 };
 
 template<typename T>
--- a/tdms/include/arrays.h
+++ b/tdms/include/arrays.h
@@ -84,6 +84,30 @@ public:
    * @param ptr Pointer to assign
    */
   void set_ptr(AxialDirection d, double *ptr);
+
+  /**
+   * @brief Determines whether all elements in the x, y, or z vector are less than a given value.
+   *
+   * @param comparison_value Value to compare elements to
+   * @param vector_length Number of elements to compare against
+   * @param component Vector to compare elements against; x, y, or z
+   * @param buffer_start Only compare elements between buffer_start (inclusive) and buffer_start+vector_length-1 (inclusive)
+   * @return true All elements are less than the comparison_value
+   * @return false At least one element is not less than the comparison_value
+   */
+  bool all_elements_less_than(double comparison_value, int vector_length,
+                                 AxialDirection component, int buffer_start = 0);
+  bool all_elements_less_than(double comparison_value, int vector_length, char component,
+                                 int buffer_start = 0);
+  /**
+   * @brief Determines whether all elements in the x, y, AND z vectors are less than a given value.
+   *
+   * @param comparison_value Value to compare elements to
+   * @param nx,ny,nz Number of elements in the nx, ny, and nz vectors respectively
+   * @return true All elements are less than the comparison_value
+   * @return false At least one element is not less than the comparison_value
+   */
+  bool all_elements_less_than(double comparison_value, int nx, int ny, int nz);
 };
 
 // TODO: docstring
--- a/tdms/include/arrays.h
+++ b/tdms/include/arrays.h
@@ -97,8 +97,6 @@ public:
    */
   bool all_elements_less_than(double comparison_value, int vector_length,
                                  AxialDirection component, int buffer_start = 0);
-  bool all_elements_less_than(double comparison_value, int vector_length, char component,
-                                 int buffer_start = 0);
   /**
    * @brief Determines whether all elements in the x, y, AND z vectors are less than a given value.
    *
--- a/tdms/include/field.h
+++ b/tdms/include/field.h
@@ -127,6 +127,16 @@ public:
     void initialise_fftw_plan(int n_threads, EHVec &eh_vec);
 
     /**
+     * @brief Fetches the largest (by absolute value) field value.
+     *
+     * Split field values are sums of the corresponding components, so this function returns the largest absolute value of the entries in
+     * (xy + xz), (yx + yz), (zx + zy)
+     *
+     * @return double Largest (by absolute value) field value
+     */
+    double largest_field_value();
+
+    /**
    * @brief Interpolates a SplitField component to the centre of a Yee cell
    *
    * @param d SplitField component to interpolate
--- a/tdms/include/field.h
+++ b/tdms/include/field.h
@@ -353,7 +353,7 @@ public:
    * @param other The other field
    * @return double Maximum relative pointwise absolute difference
    */
-  double max_pointwise_difference_over_max_element(Field &other);
+  double normalised_difference(Field &other);
 
   ~Field();
 };
--- a/tdms/include/field.h
+++ b/tdms/include/field.h
@@ -341,6 +341,20 @@ public:
    */
   void set_values_from(Field &other);
 
+  /**
+   * @brief Computes the maximum pointwise absolute difference of the other field to this one, divided by the largest absolute value of this field's components.
+   *
+   * Specifically, we compute and return the quantity
+   * $$ \frac{ \max_{i,j,k}(this[k][j][i] - other[k][j][i]) }{ \max_{i,j,k}this[k][j][i] } $$
+   * This quantity is used to determine convergence of phasors.
+   *
+   * The other field must have the same dimensions as this field.
+   *
+   * @param other The other field
+   * @return double Maximum relative pointwise absolute difference
+   */
+  double max_pointwise_difference_over_max_element(Field &other);
+
   ~Field();
 };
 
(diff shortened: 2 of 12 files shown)


tdms/include/arrays.h
```diff
@@ -172,6 +196,16 @@ class DispersiveMultiLayer{
...
+
+  /**
+   * @brief Determines whether the (background) medium is dispersive
+   *
+   * @param K_tot Number of Yee cells in the z-direction (number of entries in this->gamma)
+   * @param near_zero_tolerance Tolerance for non-zero gamma (attenuation) values
+   * @return true Background is dispersive
+   * @return false Background is not dispersive
+   */
+  bool is_dispersive(int K_tot, double near_zero_tolerance=1e-15);
```
samcunliffe, line 208: So we can have a non-dispersive dispersive multilayer? (`DispersiveMultiLayer::is_dispersive == false`)

tdms/include/field.h
```diff
@@ -126,6 +126,16 @@ class SplitField : public Grid{
      */
     void initialise_fftw_plan(int n_threads, EHVec &eh_vec);
 
+    /**
+     * @brief Fetches the largest (by absolute value) field value.
```
samcunliffe, line 130:
```suggestion
     * @brief Fetches the largest absolute value of the field.
```

tdms/include/arrays.h
samcunliffe, line 208: So we can have a non-dispersive dispersive multilayer? (`DispersiveMultiLayer::is_dispersive == false`)

tdms/include/field.h
```diff
@@ -331,6 +341,20 @@ class Field : public Grid{
    */
   void set_values_from(Field &other);
 
+  /**
+   * @brief Computes the maximum pointwise absolute difference of the other field to this one, divided by the largest absolute value of this field's components.
+   *
+   * Specifically, we compute and return the quantity
+   * $$ \frac{ \max_{i,j,k}(this[k][j][i] - other[k][j][i]) }{ \max_{i,j,k}this[k][j][i] } $$
```
samcunliffe, line 348:
```suggestion
   * \f$ \frac{ \max_{i,j,k}(this[k][j][i] - other[k][j][i]) }{ \max_{i,j,k}this[k][j][i] } \f$
```
Does that render in doxygen? I _think_ you need `\f`s.

tdms/include/field.h
```diff
@@ -331,6 +341,20 @@ class Field : public Grid{
...
+   * This quantity is used to determine convergence of phasors.
+   *
+   * The other field must have the same dimensions as this field.
+   *
+   * @param other The other field
+   * @return double Maximum relative pointwise absolute difference
+   */
+  double max_pointwise_difference_over_max_element(Field &other);
```
samcunliffe, line 356: The `@brief` docstring is nice and detailed. How about `scaled_difference` or `normalised_difference`? (Assuming either of those are accurate.)

tdms/include/grid_labels.h
```diff
@@ -11,11 +11,20 @@
...
     explicit GridLabels(const mxArray *ptr);
+
+    /**
+     * @brief Set values by copying from another GridLabels object
+     *
+     * @param other_labels The GridLabels object to copy values from
+     * @param i_l,j_l,k_l The first item to copy from the x, y, z attributes (respectively)
+     * @param i_u,j_u,k_u The final (inclusive) item to copy from the x, y, z attributes (respectively)
+     */
+    void initialise_from(GridLabels &labels_to_copy_from, int i_l, int i_u, int j_l, int j_u, int k_l, int k_u);
```
samcunliffe, line 29: Initialise != constructor?
(Just asking) why we don't have a `GridLabels` constructor?

tdms/src/arrays.cpp
```diff
@@ -27,6 +27,54 @@ void XYZVectors::set_ptr(AxialDirection d, double* ptr){
...
+    case 'y':
+      d = AxialDirection::Y;
+      break;
+    case 'z':
+      d = AxialDirection::Z;
+      break;
+    default:
+      throw runtime_error("Error - component not recognised");
+      break;
+  }
```
samcunliffe, line 68: Feels like we should be able to construct an `AxialDirection` with a `char`. Then this overloaded function becomes a one-liner (which could go in the header) and we only write the `switch` once.

tdms/src/iterator.cpp
```diff
@@ -871,10 +871,10 @@ void execute_simulation(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs
 
   //work out if we have any disperive materials
   bool is_disp = is_dispersive(materials, gamma, params.dt, I_tot, J_tot, K_tot);
-  //work out if we have conductive background
-  bool is_cond = is_conductive(rho_cond, I_tot, J_tot, K_tot);
-  //work out if we have a dispersive background
-  if (params.is_disp_ml) params.is_disp_ml = is_dispersive_ml(ml, K_tot);
+  //work out if we have conductive background: background is conductive if at least one entry exceeds 1e-15
+  bool is_cond = !(rho_cond.all_elements_less_than(1e-15, I_tot + 1, J_tot + 1, K_tot + 1));
```
samcunliffe, line 875: Would you mind renaming that symbol to `bool is_conductive`?

samcunliffe: And (for future refactorings) does it make sense to collect all of the background parameters(?) together?
Then we have a more self-evident line of code like...
```c++
bool is_conductive = Background.is_conductive(...);
```
...or whatever.

tdms/src/iterator.cpp
```diff
@@ -871,10 +871,10 @@ void execute_simulation(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs
...
   //work out if we have any disperive materials
   bool is_disp = is_dispersive(materials, gamma, params.dt, I_tot, J_tot, K_tot);
-  //work out if we have conductive background
-  bool is_cond = is_conductive(rho_cond, I_tot, J_tot, K_tot);
-  //work out if we have a dispersive background
-  if (params.is_disp_ml) params.is_disp_ml = is_dispersive_ml(ml, K_tot);
+  //work out if we have conductive background: background is conductive if at least one entry exceeds 1e-15
+  bool is_cond = !(rho_cond.all_elements_less_than(1e-15, I_tot + 1, J_tot + 1, K_tot + 1));
+  // work out if we have a dispersive background
+  if (params.is_disp_ml) params.is_disp_ml = ml.is_dispersive(K_tot);
```
samcunliffe, line 877: `ml` is the "matched" layer? So somewhere there's a `DispersiveLayer ml` declared?

Is it a pain to rename `ml` to `matched_layer`? Does it make sense?

tdms/src/iterator.cpp
```diff
@@ -4556,7 +4520,7 @@ void execute_simulation(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs
   *mxGetPr((mxArray *) plhs[25]) = maxfield;
 
   if (params.run_mode == RunMode::complete && params.exphasorsvolume) {
-    setGridLabels(input_grid_labels, output_grid_labels, E.il, E.iu, E.jl, E.ju, E.kl, E.ku);
+    output_grid_labels.initialise_from(input_grid_labels, E.il, E.iu, E.jl, E.ju, E.kl, E.ku);
```
samcunliffe, line 4523: I now maybe understand initialise != construct, but I leave my previous question just in case.

samcunliffe (approved):
👍

tdms/include/arrays.h
willGraham01: Apparently so - there's a corresponding call in `iterator.cpp`. I don't have the same feel for everything _in_ `iterator.cpp` yet like Tom did, so I'm guessing its an oddity for us to be in this case rather than common.

There's also another [`is_dispersive`](https://github.com/UCL/TDMS/blob/89fce4c9acc7e4db95d4e533dfc6e84c01a7acc0/tdms/src/iterator.cpp#L873-L875) function for the background material (which doesn't have its own class) too.

tdms/include/field.h
willGraham01: Gone with `normalised_difference` :+1:

tdms/include/grid_labels.h
willGraham01: That's a question for @t-young31 if he remembers :sweat_smile:
 
Although I think it's just because we never actually write our own labels, just cherry-pick the relevant ones from what MATLAB/our input gives us.

tdms/src/arrays.cpp
willGraham01: It's actually [already constructed](https://github.com/UCL/TDMS/blob/89fce4c9acc7e4db95d4e533dfc6e84c01a7acc0/tdms/include/globals.h#L42-L47) like this - I didn't realise!

tdms/src/iterator.cpp
willGraham01: [Right here](https://github.com/UCL/TDMS/blob/wgraham-refactor_iterator_3/tdms/src/iterator.cpp#L438) approx 400 lines ago.

Have run a VSCode refactor over the ml symbol

PR #205, opened 2022-12-19
SurfacePhasors class

willGraham01:
Continues #63 |

## `SurfacePhasor` class

Adds a class that handles the complex amplitude extraction across the user-specific surface in `iterator.cpp`. 

Removes a data clump (`surface_EH{r,i}`, `surface_vertices`, `vertex_list` and other associated variables) and groups together what were previously a collection of standalone functions.

Also adds a destructor for this class so memory cleanup is automatic, rather than having to be called manually around line 4620. 

### Extensions

I suspect either this class or a very similar one can also be used to remove the data clump associated to the complex amplitude extraction at the user-specified vertices. We may even be able to directly generalise the `SurfacePhasor` class into a `PhasorExtraction` super-class or something...


tdms/include/surface_phasors.h
```diff
@@ -0,0 +1,122 @@
...
+   * @param _f_ex_vector_size The FrequencyExtractionVector size, which we need to reserve sufficient memory
+   */
+  void set_from_matlab_array(mxArray *mx_surface_vertices, int f_ex_vector_size);
+
+  /**
+   * @brief Zeros the surface_EH{r,i} arrays
+   */
+  void zero_surface_EH() {
+    for (int k = 0; k < f_ex_vector_size; k++) {
+      for(int j = 0; j < 6; j++) {
```
samcunliffe, line 47:
```suggestion
      for(int j = 0; j < 6; j++) { // 6 field components: Ex, y, z, and Hx, y, z
```
Just because I didn't immediately remember your indexing comment.

tdms/src/surface_phasors.cpp
```diff
@@ -0,0 +1,189 @@
...
+  // set the number of vertices on the surface using the array dimensions
+  n_surface_vertices = dimensions_pointer_out[0];
+
+  //cast the vertex array as a 2-d integer array
+  surface_vertices = cast_matlab_2D_array((int *) mxGetPr((mxArray *) mx_surface_vertices),
+                                          dimensions_pointer_out[0], dimensions_pointer_out[1]);
+
+  //create space for the complex amplitudes E and H around the surface. These will be in a large complex
+  //array with each line being of the form Re(Ex) Im(Ex) Re(Ey) ... Im(Hz). Each line corresponds to the
+  //the vertex with the same line as in surface_phasors.surface_vertices
```
samcunliffe, line 28:
```suggestion
  // Create space for the complex amplitudes E and H around the surface. These will be in a large complex
  // array with each line being of the form Re(Ex) Im(Ex) Re(Ey) ... Im(Hz). Each line corresponds to the
  // the vertex with the same line as in surface_phasors.surface_vertices.
```

tdms/src/surface_phasors.cpp
```diff
@@ -0,0 +1,189 @@
...
+  complex<double> phaseTermE, phaseTermH, cphaseTermE, cphaseTermH;
+
+  phaseTermE = fmod(omega * ((double) n) * params.dt, 2 * DCPI);
+  phaseTermH = fmod(omega * ((double) n + 0.5) * params.dt, 2 * DCPI);
+
+  cphaseTermH = exp(phaseTermH * IMAGINARY_UNIT) * 1. / ((double) Nt);
+  cphaseTermE = exp(phaseTermE * IMAGINARY_UNIT) * 1. / ((double) Nt);
+
+  //loop over every vertex in the list
+#pragma omp parallel default(shared) private(F, phaseTermE, phaseTermH, vindex)
```
samcunliffe, line 79: I don't have a specific suggestion for here, but in general, perhaps we should add comments about exactly how this is parallelised. I mean like: the parallelization strategy.

```c++
// here we can parallelise the calculation of foo as we loop over bar
```

samcunliffe: Also: did you test locally that with `OMP_NUM_THREADS=8` (or whatever) that we still pass all of the system tests?

Yes, I'm asking you to do a manual test 👎 . It's probably my fault because I didn't get to #89 yet.

tdms/src/surface_phasors.cpp
```diff
@@ -0,0 +1,189 @@
...
+  surface_EHi[frequency_index][1][vertex_index] += imag(F.Ey);
+  //Hy
+  surface_EHr[frequency_index][4][vertex_index] += real(F.Hy);
+  surface_EHi[frequency_index][4][vertex_index] += imag(F.Hy);
+  //Ez
+  surface_EHr[frequency_index][2][vertex_index] += real(F.Ez);
+  surface_EHi[frequency_index][2][vertex_index] += imag(F.Ez);
+  // Hz
+  surface_EHr[frequency_index][5][vertex_index] += real(F.Hz);
+  surface_EHi[frequency_index][5][vertex_index] += imag(F.Hz);
```
samcunliffe, line 176: Don't have a suggestion for how to improve, but this _must_ be loop-able.
Also: 0, 3, 1, 4, 2, 5 does not respect my understanding of the ordinality of numbers 😛 .

tdms/tests/unit/array_tests/test_FullFieldSnapshot.cpp
```diff
@@ -0,0 +1,45 @@
+/**
+ * @file test_FullFieldSnapshot.cpp
+ * @author William Graham (ccaegra@ucl.ac.uk)
+ * @brief Unit tests for CollectionBase class and its subclasses (CCollection, DCollection)
```
samcunliffe, line 4: The `@brief` comment is a lie, no?

samcunliffe (approved):
Mostly I want more doc comments from you, old bean.

tdms/src/surface_phasors.cpp
willGraham01: Have run these locally and looks like we're all good. Notable difference in speed of course (as expected), but the tests don't complain.

Issue #87, opened 2022-09-09
Use the new logging


samcunliffe:
[Low priority]

Now we have spdlog we should use it. (A few `cerr`s should be `spdlog::error` and even `spdlog::info`.)

PR #200, opened 2022-12-13
Rename phasorinc, add {C,D}Collection documentation, and `main()` → `main.cpp`

samcunliffe:
There was a hanging documentation and a renaming task from our last meeting with Peter.

I also include a few minor sanity fixes: 
* I moved `main` into `main.cpp`.
* Fixed the broken `<details>` tags (or at least it's fixed with my local doxygen version).
* Updated the version of doxygen (I hope).

PR #206, opened 2022-12-20
PRHS into a readable class

willGraham01:
Moving towards #63 |

The argument `prhs` to `execute_simulation` is rather unintuitive, being a hangover from when `mexfunction` was used. With regards to moving towards #70, and making the codebase a bit more readable, this object is broken out into a new class `InputMatrices`. This class steals some of the functionality of `openandorder`, and is designed to make the "unpacking" part of `iterator.cpp` more readable.

In theory, we can also do away with the intermediate variables that are defined in `iterator.cpp` and work directly with `prhs` now (although that is not recommended for readability purposes).

- Constant increments of an `input_counter++` are no longer needed when unpacking the input matrices - they can be accessed by passing in their _names_.
- Validation that the input arrays are of the correct type/shape are performed when they are read in by the `InputMatrices` class, rather than at the start of `execute_simulation`. This avoids potentially redundant setup/memory allocation.
- The constants like `omega`, `Nt`, etc stored in `SimulationParameters` are now set by calling the `unpack_from_input_matrices` method, rather than being done directly inside `execute_simulation`, reducing the code bloat.
- A number of `fprintf`s -> `spdlog::info` in line with #87.
- General Docstring tidy up in files that are touched by this PR.


tdms/include/input_output_names.h
```diff
@@ -0,0 +1,32 @@
...
+    extern const char *matrixnames_gridfile[1];
+    // oall output matrices we might want to write
+    extern const char *outputmatrices_all[NOUTMATRICES_WRITE_ALL];
+    // output matrices we want to write in a compressed (-m) output
+    extern const char *outputmatrices[NOUTMATRICES_WRITE];
+    // indices of the matrices to save when saving a complete system
+    extern int matricestosave_all[NOUTMATRICES_WRITE_ALL];
+    // indices of the matrices to save when saving a compressed output
+    extern int matricestosave[NOUTMATRICES_WRITE];
+}
```
samcunliffe, line 32: Ah, this is likely going to clash with my refactoring/reworking. (You weren't to know because I didn't tell you or push to my draft PR).

I went for `std::vector<std::string>` (which could be defined as const and live in a namespace). My reasoning: a vector knows its own `size_t` and one can do much nicer-looking loops like:
```c++
for (auto matrixname : matrixnames) {
     file.read(matrixname);
}
```

samcunliffe: Come to think of it, the matrices in our input and output files _define_ our file type. So perhaps a `tdms::File` is a HDF5 of this format.

tdms/include/simulation_parameters.h
```diff
@@ -114,4 +127,11 @@ class SimulationParameters{
     void set_spacing_stride(const double* vector);
```
samcunliffe, line 127: Weird. Did I not write a docstring!?

tdms/src/input_matrices.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+using namespace std;
+using namespace tdms_matrix_names;
+
+int index_from_matrix_name(const char *matrix_name) {
+    for (int i = 0; i < NMATRICES; i++) {
+        if(are_equal(matrix_name, matrixnames[i])) { return i; }
+    }
+    throw runtime_error("Matrix name " + string(matrix_name) + " not found in matrixnames.");
+}
```
samcunliffe, line 20: An `std::vector` can be searched with [`std::find`](https://en.cppreference.com/w/cpp/algorithm/find)... doesn't make the code spectacularly more beautiful but _you_ don't need to play with string equality.

```c++
int index_from_matrix_name(const std::string& matrix_name) {
    auto position = std::find(matrixnames.begin(), matrixnames.end(), matrix_name);
    if (position == matrixnames.end()) throw runtime_error(...);
    return std::distance(matrixnames.begin(), position);
}
```

tdms/src/input_output_names.cpp
```diff
@@ -0,0 +1,14 @@
...
+    const char *matrixnames_gridfile[1] = {"fdtdgrid"};
+    const char *outputmatrices_all[NOUTMATRICES_WRITE_ALL] = {"Ex_out","Ey_out","Ez_out","Hx_out","Hy_out","Hz_out","x_out","y_out","z_out","Ex_i","Ey_i","Ez_i","Hx_i","Hy_i","Hz_i","x_i","y_i","z_i","vertices","camplitudes","facets","maxresfield","Id","fieldsample","campssample"};
+    const char *outputmatrices[NOUTMATRICES_WRITE] = {"Ex_out","Ey_out","Ez_out","Hx_out","Hy_out","Hz_out","x_out","y_out","z_out","Ex_i","Ey_i","Ez_i","Hx_i","Hy_i","Hz_i","x_i","y_i","z_i","camplitudes","maxresfield","Id","fieldsample","campssample"};
+    int matricestosave_all[NOUTMATRICES_WRITE_ALL] = {0,  1,  2,  3,  4,  5,  10, 11, 12,
+                                                      13, 14, 15, 16, 17, 18, 19, 20, 21,
+                                                      22, 23, 24, 25, 26, 27, 28};
+    int matricestosave[NOUTMATRICES_WRITE] = {0,  1,  2,  3,  4,  5,  10, 11, 12, 13, 14, 15,
+                                              16, 17, 18, 19, 20, 21, 23, 25, 26, 27, 28};
+}
```
samcunliffe, line 14: Let me `cherry-pick` my changes over to separate branch from main so I can show you what I did and see if you like it.

tdms/include/simulation_parameters.h
samcunliffe: 😱 You've got to watch this guy, [he'll try to get away with schenanegans](https://github.com/UCL/TDMS/pull/200/files).

samcunliffe, line 127:
```suggestion
    /**
     * @brief Set the surface spacing stride.
     * The x, y, z step size for extracting phasors (in matlab this is called 'phasorinc')
     * @param vector the x, y, z steps (i.e. x = 2 means extract from every 2nd Yee cell.)
     */
    void set_spacing_stride(const double* vector);
```

tdms/src/input_output_names.cpp
samcunliffe: [Here's](https://github.com/UCL/TDMS/compare/matrixnames-as-std-vectors) a commit (which probably doesn't compile), but you get the idea.

Whaddya think?

willGraham01: A `vector` of `string`s does sound more appealing than the const char stuff we have floating around currently.

I'll change the `matrixnames` and the like accordingly, but I'll keep them in the new file structure (header + source file defining the matrix names in a namespace).

samcunliffe (approved):

samcunliffe:
Looks all good!
When you're happy to merge, could you reject #207 at the same time, please?

PR #207, opened 2022-12-21
`matrixnames` as `std::vector<std::string>`s

samcunliffe:
I don't think we want to merge this, given #206. Instead, let me leave it here so you can `cherry-pick` or `checkout` certain files or whatever. Assuming that's helpful.

### DIFF
--- a/tdms/src/matrix_collection.cpp
+++ b/tdms/src/matrix_collection.cpp
@@ -3,6 +3,8 @@
 #include <string>
 #include <stdexcept>
 
+#include <spdlog/spdlog.h>
+
 #include "mat_io.h"
 
 using namespace std;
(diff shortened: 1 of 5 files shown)


tdms/src/openandorder.cpp
```diff
@@ -29,7 +31,7 @@ void openandorder(const char *mat_filename, char **matrix_names, const mxArray *
   //fprintf(stderr, "Got all %d matrices\n",nmatrices);
   //fprintf(stderr, "%d\n",mxGetFieldNumber( (mxArray *)matrixptrs[0], "I_tot"));
 
-  if (!are_equal(matrix_names[0], "fdtdgrid")){
+  if (matrix_names[0] != "fdtdgrid"){
```
samcunliffe, line 34: Ah, of course this doesn't work if `"fdtdgrid"` is the _last_ entry in the vector.
I'm going to leave this up to confirm the build fails before fixing.

samcunliffe: [Very good](https://github.com/UCL/TDMS/actions/runs/3750619750/jobs/6370566162).

tdms/src/matrix_collection.cpp
```diff
@@ -39,5 +42,8 @@ MatFileMatrixCollection::~MatFileMatrixCollection() {
   if (mat_file != nullptr){
     matClose(mat_file);
   }
-  mxFree(matrix_names);
+  std::vector<char*> matrix_names_cstrs{};
```
willGraham01, line 45: Can I ask what's going on here? Why can't we just
```c++
matrix_names.clear()
```
to deallocate the memory?

samcunliffe: Ah, no it's just me converting the `std::vector<std::string>` back to an array of char* (via an `std::vector<char*>`). Because that's what `mxFree` wants.

samcunliffe: Oh wait, that's what [`mxFree` does](https://uk.mathworks.com/help/matlab/apiref/mxfree.html). I get your question. 


The answer isL code-blindness. I translated without thinking.

Can probably delete all of this and leave the vector to go out of scope on class destruction.

willGraham01:
Closing as per changes in #206. [Comment](https://github.com/UCL/TDMS/pull/206#issuecomment-1362776252)

PR #209, opened 2023-01-04
Vertex Phasors Class

willGraham01:
Continues #63 |

Along similar lines to #205, breaks out extraction of the phasors at the user-specified vertices into its own class.

### DIFF
--- a/tdms/include/fieldsample.h
+++ b/tdms/include/fieldsample.h
@@ -37,7 +37,7 @@ public:
    * @param pml A description of the perfectly matched layer being used in this simulation
    * @param n_simulation_timesteps The (total) number of timesteps in this simulation
    */
-  void extract(ElectricSplitField E_split, PerfectlyMatchedLayer pml, int n_simulation_timesteps);
+  void extract(ElectricSplitField &E_split, PerfectlyMatchedLayer &pml, int n_simulation_timesteps);
 
   ~FieldSample();
 };
--- a/tdms/include/fieldsample.h
+++ b/tdms/include/fieldsample.h
@@ -1,0 +1,43 @@
+/**
+ * @file fieldsample.h
+ * @brief Class that stores the extracted field values at user-defined vertices.
+ */
+#pragma once
+
+#include "arrays.h"
+#include "field.h"
+#include "simulation_parameters.h"
+
+class FieldSample{
+
+private:
+  double**** tensor = nullptr;
+
+public:
+  mxArray* mx;       // Matlab array
+
+  Vector<int> i;     // Indices along the x-direction of locations at which to sample the field
+  Vector<int> j;     // Indices along the y-direction of locations at which to sample the field
+  Vector<int> k;     // Indices along the z-direction of locations at which to sample the field
+  Vector<double> n;  // Vector of the moments of the field to sample
+
+  explicit FieldSample(const mxArray *ptr);
+
+  // Return true if all vectors in this instance are non-empty (have size > 0)
+  bool all_vectors_are_non_empty() const{
+          return i.size() > 0 && j.size() > 0 && k.size() > 0 && n.size() > 0;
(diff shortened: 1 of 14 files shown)


tdms/include/fieldsample.h
```diff
@@ -0,0 +1,43 @@
...
+private:
+  double**** tensor = nullptr;
+
+public:
+  mxArray* mx;       // Matlab array
+
+  Vector<int> i;     // Indices along the x-direction of locations at which to sample the field
+  Vector<int> j;     // Indices along the y-direction of locations at which to sample the field
+  Vector<int> k;     // Indices along the z-direction of locations at which to sample the field
+  Vector<double> n;  // Vector of the moments of the field to sample
```
samcunliffe, line 22:
```suggestion
  mxArray* mx;       // Matlab array

  Vector<int> i;     //< Indices along the x-direction of locations at which to sample the field
  Vector<int> j;     //< Indices along the y-direction of locations at which to sample the field
  Vector<int> k;     //< Indices along the z-direction of locations at which to sample the field
  Vector<double> n;  //< Vector of the moments of the field to sample
```

tdms/include/fieldsample.h
```diff
@@ -0,0 +1,43 @@
...
+
+  explicit FieldSample(const mxArray *ptr);
+
+  // Return true if all vectors in this instance are non-empty (have size > 0)
```
samcunliffe, line 26:
```suggestion
  /** Return true if all vectors in this instance are non-empty (have size > 0) */
```
(Keep doxygen happy)

tdms/src/fieldsample.cpp
```diff
@@ -0,0 +1,58 @@
...
+    mx = mxCreateNumericArray(n_dims, (const mwSize *) dims, mxDOUBLE_CLASS, mxREAL);
+  }
+}
+
+void FieldSample::extract(ElectricSplitField &E_split, PerfectlyMatchedLayer &pml,
+                          int n_simulation_timesteps) {
+  double Ex_temp = 0., Ey_temp = 0., Ez_temp = 0.;
+
+/* Extract the (electric) field at each of the vertices.
+Since the split-field has already been computed, we can do this in parallel by reading the values from the split field and interpolating to the vertices independently of each other.
```
samcunliffe, line 29: 👍 👍

samcunliffe (approved):
LGTM.

My only really comment: I wonder if we should have an abstract `Phasors` class. Can you factor out any common functionality? Does it buy us anything?

tdms/include/vertex_phasors.h
```diff
@@ -0,0 +1,105 @@
...
+#include "arrays.h"
+#include "field.h"
+#include "grid_labels.h"
+
+/**
+ * Class container for handling complex amplitude samples at the vertices, and their extraction.
+ *
+ * Abbreviated to CAmpSample in MATLAB code
+ */
+class VertexPhasors {
```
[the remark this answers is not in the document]
willGraham01: I gave this an afternoon when developing, and concluded no.

Despite extracting the same thing; the method by which this happens, the data structures we need to setup to store the phasors, and the information that we need to take from the input file to know what we're doing, are considerably different in both cases. 

The only common functionality I could find was the correspondence between the `surface_EH{r,i}` and `camplitudes{R,I}` variables, and even these are different shapes since the `vertex_phasors` class does not always extract all the field components whilst the `surface_phasors` class appears to do so.

That's not to say it's _impossible_ for us to do this, just I don't think it's _worth_ us doing this. At least whilst we're this side of #70 (as MATLAB gone -> C++ native structures everywhere -> might open up common functionality / data storage options)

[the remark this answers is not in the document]
samcunliffe: Gottit. Fine by me.

PR #210, opened 2023-01-05
Iterator Class Breakdown

willGraham01:
# [TLDR] Major Changes to `Iterator.cpp`
Continues #63 | [Pros](#Pros) | [Cons](#Cons) | [What does any of this actually do, Will?](#what-does-what)

- Introduces several `Iterator_` classes and subclasses
- `Iterator` is now bottom-level class, `iterator.h` contains declarations for said class
- Supporting classes handle data chunks that can go in/out of scope together
- `execute_simulation` is reduced to a series of calls to `Iterator` methods, and some interaction between the output `plhs` and `Iterator` attributes.
- `iterator.cpp` itself is gone. A new folder `src/iterator/` contains the new source files. Similarly, `include/iterator/` contains the header files.
- `execute_simulation` is declared in the `iterator.h` header, but is not a method of the `Iterator` class since it involves `InputMatrices`, `plhs`, `Iterator`, and the command-line flags.

Current reorganisation is the following:
Untitled-2022-12-13-1412

## <a href="#user-content-pros" id="pros">Pros</a>
- Refactoring the main loop no longer requires huge data clumps being passed into subfunctions. All the variables you need at that point are class attributes (or have methods to fetch them).
- `execute_simulation` can be read as a series of steps
- Variables are only in scope _when they are required_. This is achieved through only declaring class members that are necessary for the functionality of that class, and postponing any further declarations to a "lower point" in the class hierarchy. An exception to this is the `Iterator_PreLoopVariables` class, which is essentially just the declarations we need for the main loop so that the source files are readable.
- Classes handle memory free-ing from MATLAB and malloc'd blocks more effectively, and we can compliment calls to these routines with the corresponding `free`s in the class destructors.
- Provides distinction between the setup for the main loop (unpacking variables and pre-loop variable settings); the main loop execution, post-main-loop analysis, and assignment to the outputs.
- Highlights the need for an `output_matrices` class to tear down the MATLAB memory that `plhs` holds, but never frees.
- Adds some docstrings to variables whose role is now clear from context.

## <a href="#user-content-cons" id="cons">Cons</a>
- `run_main_loop` is still ~3000+ lines of uncommented code (but at least it can be broken down further without too much issue)
- MATLAB's `mxFree` and `mxDestroyArray` are not interchangeable... and getting them wrong will cause seg-faults and other nasty things to happen.

#### `Iterator_IndependentObjectsFromInfile`
- Creates the data structures that store input data from `InputMatrices`.
- Only creates variables/arrays that can be read directly from the input file, and variables that are immediately derived from them (`I_tot, J_tot, K_tot` in particular).

#### `Iterator_ObjectsFromInfile`
Some of the objects from the input file can only be properly setup in `C++` arrays if we have loaded _other_ inputs from the input file first. Hence, we inherit the `IndependentObjectsFromInfile` class, construct everything in there first, then load in the rest of the data.
- Loads in remaining arrays that need extra help to get setup.
- Also defines a few parameters that are immediately set from this import.

#### `Iterator_PreLoopDeclarations`
Objects that are needed in the main loop, but also afterwards in the writing to outputs (or writing to the addresses the output will read from during the main loop).
- Class has no constructor, and no methods. Members are default-initialised and can handle their own tear-down.
- Serves as a nice container so that lower-down classes do not get bloated, and as a point of organisation for variable scoping.

#### `Iterator_Timers`
Handles the timing of the main loop and (optional) inner-loop timings too.
- Timers are self-contained and don't interact with any other variables, but their scope needs to be both inside and outside the main loop.

#### `Iterator_Executor`
- Handles the main loop
- Handles variables whose scope is limited to the main loop
- Updates necessary variables at each iteration of the main loop

#### `Iterator`
Adds methods to interact with the output pointers, `plhs`, that can be called from within the `execute_simulation` method.
- Handles post-main loop functionality

#### `execute_simulation`
Does what it did beforehand. But now you can read it as a series of instructions and references to the classes we created above. Also it's now declared in `iterator.h` because reasons.


willGraham01:
Superceeded by #215

PR #213, opened 2023-01-12
`Output_Matrices class`

willGraham01:
# `OutputMatrices` object

This object handles the data structures that contain output data, the writing of this data to the output file, and the tear-down of any allocated memory (*cough* MATLAB *cough*) that the outputs have reserved after writing.

The class itself only has one member, the `matrix_pointers` array that containers the pointers to the output data that is to be written. This will likely change when #70 completes, and we can assign our own `arrays.h` classes as members of this class instead.

## Other random notes:
- Found that `extractPhasorsPlane` writes to arrays that are never referenced again, and do not constitute outputs. Have left this code in for now, and flagged this in an issue #214.
- Removed `plhs` entirely - `OutputMatrices` is now the go-to. It's also an _output_ of `execute_simulation` rather than passed in by reference.
- Removes the `nlhs` and `nrhs` checks in `execute_simulation`. These were for when we were using a `mexFunction` and serve no purpose now that we have classes that can tell us how many inputs/outputs they are managing.
- `openandorder.cpp` is no more - functionality has been absorbed into other classes


tdms/src/iterator.cpp
```diff
@@ -4582,21 +4584,21 @@ void extractPhasorsPlane(Matrix<complex<double>> &iwave_lEx_bs, Matrix<complex<d
       complex<double> subResult = 0.;
       //Eyz
       subResult = (E.yz[K1][j][i] + E.yx[K1][j][i]) * exp(phaseTerm * IMAGINARY_UNIT) * 1. / ((double) Nt);
-      iwave_lEy_bs[j][i] += subResult;
+      iwave_lEy_bs[i][j] += subResult;
       //Hxz
       subResult = (H.xz[K1 - 1][j][i] + H.xy[K1][j][i]) * exp(phaseTerm * IMAGINARY_UNIT) * 1. / ((double) Nt);
-      iwave_lHx_bs[j][i] += subResult;
+      iwave_lHx_bs[i][j] += subResult;
```
samcunliffe, line 4453: 🐛 ?

samcunliffe (commented):
All looks great to me. I can't really spot anything.

Does it make sense to (unit) test this? _I actually think maybe not, since it's all code for getting matrices out of the file: which is covered by the system tests._

[But always worth asking.]

samcunliffe (approved):

willGraham01:
> Does it make sense to (unit) test this? _I actually think maybe not, since it's all code for getting matrices out of the file: which is covered by the system tests._

I agree that this behaviour that should be captured by the system tests. Plus I don't like the idea of having to manually generate 25+ MATLAB objects in a unit test and then read in the corresponding file!

Furthermore, we already test that the `arrays.h` and `field.h` objects _can be linked_ to a MATLAB pointer in the unit tests, and that's all we're doing to interact with them here. In the upcoming refactor #215 I'm moving these objects (that are linked to the pointers) into the scope of the `OutputMatrices` class to make this much cleaner.

PR #218, opened 2023-01-20
Refactor the Main Loop

willGraham01:
Continues #63 |

- Breaks out `check_phasor_convergence` method at the start of the loop.
- Breaks out `compute_detector_functions` method, which takes with it some variables that can be scoped to this process alone.
- Breaks out `new_acquisition_period` method, which zeros angular norms and normalisation factors when a new period of harmonic illumination begins.
- Breaks out `end_of_iteration_steps` method, which handles some small tasks pertaining to the setup of the next iteration upon completion of the present.
- Breaks out `update_Exy` and `update_Exz` methods - these are the update equations for the corresponding split-field components. Given sufficient time, it should be possible to extract a method for all of the split-field components, and potentially even generalise it so that we don't have nearly-identical code repeated 6 times in the repository. For now at least, it is removed from the main loop to make the file slightly more readable.


tdms/include/field.h
```diff
@@ -308,6 +308,21 @@ class Field : public Grid {
   void set_phasors(SplitField &F, int n, double omega, double dt, int Nt);
 
   // TODO: Docstring
+  /**
+   * @brief Compute the phasor_norm of the current field and add it to the
+   * current norm-value.
+   *
+   * Note that the calls:
```
samcunliffe, line 315:
```suggestion
   * @note the calls:
```

tdms/src/simulation_manager/execute_phasor_subfunctions.cpp
```diff
@@ -0,0 +1,148 @@
...
+      (inputs.params.run_mode == RunMode::complete) &&
+      (inputs.params.source_mode == SourceMode::steadystate) &&
+      inputs.params.exphasorsvolume) {
+
+    dft_counter = 0;
+
+    double tol = outputs.E.normalised_difference(E_copy);
+    if (tol < TOL) { return true; }// required accuracy obtained
+
+    spdlog::debug("Phasor convergence: {} (actual) > {} (required)", tol, TOL);
```
samcunliffe, line 34: I might have had a separate purge function (to clarify the higher-level logic).

samcunliffe (approved):

PR #222, opened 2023-01-26
Add clang-format to the pre-commit.

willGraham01:
:1st_place_medal: Here's every single warning: [pre-commit-on-all.txt](https://github.com/UCL/TDMS/files/10511267/pre-commit-on-all.txt)

#### `linting.yaml`
Workflow renamed to "Run code linting" (previously "Run python code linting").

#### `.pre-commit-config.yaml`

Invokes [pocc/pre-commit-hooks](https://github.com/pocc/pre-commit-hooks) to force `C++` linting:
- `clang-format` is pointed at our [`.clang-format`](https://github.com/UCL/TDMS/blob/wgraham-cpp-linter/.clang-format)

#### Future considerations:
We looked at a bunch of other pre-commit tools. Here's some stuff about them that we ended up not including in this PR but might be wanted going forward.

- `clang-tidy` can be pointed to our compile commands. This requires `compile_commands.json` to be present. Can be done by passing `DCMAKE_EXPORT_COMPILE_COMMANDS=ON` when invoking `cmake`.
- `cpplint` filters `whitespace/comments` due to a conflict with `Doxygen` and `clang-tidy` standards. Suppresses `legal/copyright` because our developer docs don't require a copyright notice in header files. Runs quietly so we don't know its there if it has nothing to say.

PR #223, opened 2023-01-31
`clang-format` all the things!

samcunliffe:
One step towards a consistent style in the codebase. This is `clang-format` run over the C++ portion of the code as according to the rules in `.clang-format`. I've (either temporarily or permanently) turned off `cpplint`, `cppcheck`, and `clang-tidy`. The latter I would advocate we drop. The former two we can reenable over on #222 if we also fix the `cppcheck` recommendations.

## Tweaks in `.clang-format` w.r.t what @t-young31 had

* 80 characters line length.
* Wrap long comments please.
* Don't indent namespaces.

[Other rules](https://clang.llvm.org/docs/ClangFormatStyleOptions.html): it would be nice to check we're happy with them. We can also remove specific rules from our `.clang-format` and it will just default to whatever Google says.

## MATLAB

`.m` scripts are treated as objective C by `clang-format` (clash in file extension conventions). The only way I've managed to ignore these is by putting `.clang-format`'s in the directories of MATLAB scripts. There doesn't seem to be an "ignore these file extensions" setting.

## TODOs (done!)

* [x] Long comments.
* [x] Namespace indentation.
* ~~Can we run this over MATLAB code?~~


tdms/include/globals.h
```diff
@@ -55,29 +49,27 @@ enum FieldComponents { Ex = 1, Ey, Ez, Hx, Hy, Hz };
...
 // **************************************
 
-namespace tdms_phys_constants
-{
-const double EPSILON0 = 8.85400e-12;                         // free space electric permitivity
-const double MU0 = 4.0 * tdms_math_constants::DCPI * 1.0e-7; // free space magnetic permeability
-const double LIGHT_V = 1.0 / sqrt(EPSILON0 * MU0);           // free space light velocity
-const double Z0 = 376.734;                                   // free space inpedance
-}
+namespace tdms_phys_constants {
```
samcunliffe, line 70: Namespace indentation? 
Check with Google style and add something to `.clang-format`.

tdms/include/interpolation_methods.h
```diff
@@ -24,64 +24,66 @@
...
+  BAND_LIMITED_1 = 6,// use bandlimited interpolation w/ interp position = 1
+  BAND_LIMITED_2 = 8,// use bandlimited interpolation w/ interp position = 2
+  BAND_LIMITED_3 =
+          9,// use bandlimited interpolation w/ interp position = 3 [Preferred method if available]
+  BAND_LIMITED_4 = 7,// use bandlimited interpolation w/ interp position = 4
+  BAND_LIMITED_5 = 5,// use bandlimited interpolation w/ interp position = 5
+  BAND_LIMITED_6 = 3,// use bandlimited interpolation w/ interp position = 6
+  BAND_LIMITED_7 = 2,// use bandlimited interpolation w/ interp position = 7
+  BAND_LIMITED_CELL_ZERO =
+          1,// use bandlimited interpolation to interpolate to the centre of Yee cell 0
```
samcunliffe, line 38: (Soft opinion) I want to place it above rather than split the line. Should play well with doxygen too.

tdms/include/fieldsample.h
```diff
@@ -8,43 +8,51 @@
...
 public:
-  mxArray* mx;       // Matlab array
+  mxArray *mx;// Matlab array
 
-  Vector<int> i;     //< Indices along the x-direction of locations at which to sample the field
-  Vector<int> j;     //< Indices along the y-direction of locations at which to sample the field
-  Vector<int> k;     //< Indices along the z-direction of locations at which to sample the field
-  Vector<double> n;  //< Vector of the moments of the field to sample
+  Vector<int> i;   //< Indices along the x-direction of locations at which to
+                   // sample the field
```
samcunliffe, line 20: Another TODO:
- [ ] Check doxygen correctly renders this (!!!)

willGraham01: Doxygen will, however you're documenting a class member so you need the `!<` syntax rather than just `<`:
```
Vector<int> i; //!< Indices along the x-direction of locations at which to 
               //!< sample the field
```

willGraham01 (approved):
There was [this](https://github.com/UCL/TDMS/pull/223#discussion_r1094298979) concerning Doxygen, but I suggest that we leave policing these things to #222.

Other than that, looks good... at least the `.clang-format` does, and the ~15 `.h/.cpp` files I cherry-picked to examine seem to be consistent with it. So I have faith in the rest of the process...

tdms/include/fieldsample.h
samcunliffe: Cool. Confirmed. `clang-format` respects the doxygen comments (but the doxygen comment syntax itself needs to be correct).

PR #227, opened 2023-02-06
Remove CLI interpolation method selection - now only read from file. …

### LINKED ISSUE #217: Update to `iteratefdtd_matrix`: Command line options moved to input file
Subtask of #208 | Blocked by #240 

### `usecd` and `intmethod` in the input file

These changes are going to touch multiple files at once, _and_ require that we update the system test framework since we no longer need to read in command-line options.
- [ ] The `InputMatrices` class needs to be able to read in the `usecd` and `intmethod` variables, and to handle when one may not be present in the input file. `ObjectsFromInfile` sets the corresponding `enums` accordingly. `ArgumentSpace` looses command-line functionality to set the solver method and interpolation methods. Reporting the values of these options should moved from the `main()` body into the `SimulationManager` class.
- [ ] `SimulationManager` can be initialised with less arguments if `InputMatrices` now reads the aforementioned attributes.
- [ ] The default interpolation method is currently band-limited, however the default needs to change to cubic. Also, there are currently two `enums` that encode the interpolation method that is to be used - these need to be reconciled.
- [ ] The default solver method is PSTD but it should be FDTD. Again, `enum` cleanup might be necessary.
- [ ] The `test_sytem.py` script needs to be altered, since it is no longer necessary to specify command-line options in the tests. `test_regen.py` similarly.
- [ ] Implement #261
- [ ] Zenodo uploads should have the "redundant" flags removed - this might result in another need to re-upload.

The following tasks are closed with #253:
- [x] The new `iteratefdtd_matrix.m` needs to be added to the repository
- [x] ~Tests 17 and 18 need their generation methods added to the repository. `test_regen` and the associated generation scripts need updating too~
- [x] `arc_19` should be absorbed into the MATLAB unit testing framework
- [x] Test files need to be updated and re-uploaded to Zenodo. In particular, old tests need to have the `intmethod` and `usecd` options set properly. 
- [x] MATLAB tests need to be updated as `iteratefdtd_matrix` is one of the few functions we test

## To resolve

- [x] Consistency check between Peter's version of TDMS and the version on `main`, for the reference data in `arc_17` and `arc_18`. First, these tests need to have their reference data regenerated in accordance with the `tdms` version on main, using cubic interpolation, _and the old version of `iteratefdtd_matrix`. We should only see BLi-cubic discrepancies in the `Ex_i` outputs due to the older versions of `tdms` using a mixture of BLi and cubic schemes. Then, this can be taken as the reference outputs for the inputs to these tests using the new version of `iteratefdtd_matrix`.

willGraham01:
Closes #217 |

### `usecd` and `intmethod` in the input file

These changes are going to touch multiple files at once, _and_ require that we update the system test framework since we no longer need to read in command-line options.
- [ ] The `InputMatrices` class needs to be able to read in the `usecd` and `intmethod` variables and set the corresponding `enums` accordingly. `ArgumentSpace` looses this functionality. Reporting the values of these options should be moved from the `main()` body into the `InputMatrices` class.
- [ ] `SimulationManager` can be initialised with less arguments if `InputMatrices` now reads the aforementioned attributes (subject to [PR215](https://github.com/UCL/TDMS/pull/215)).
- [ ] The default interpolation method is currently band-limited, however the default needs to change to cubic. Also, there are currently two `enums` that encode the interpolation method that is to be used - these need to be reconciled.
- [ ] The default solver method is PSTD but it should be FDTD. Again, `enum` cleanup might be necessary.
- [ ] Test files need to be updated and re-uploaded to Zenodo. In addition, the `test_sytem.py` script needs to be altered, since it is no longer necessary to specify command-line options in the tests.
- [ ] The new tests need to be uploaded to Zenodo (and we should decide whether or not we are using cubic or band-limited interpolation). Additionally, we might want Peter to run his reference inputs through the current version of `tdms`, where the interpolation method is consistent (as older versions used a mix of BLi and cubic).
- [ ] Update test framework.
- [ ] Add new tests 17 and 18


willGraham01:
Closing this since we need a consistent way of re-generating the system tests before we progress with this. New PR will be opened following #229

Issue #208, opened 2022-12-22
Update to `iteratefdtd_matrix`


prmunro:
I have made the updates to iteratefdtd_matrix, which will change the mat file which is used as input to TDMS (for my reference, iteratefdtd_matrix was modified in in distribute_arc_01/newinput). The modified file is attached. Here is a description of the changes. Test cases have also been uploaded. 
[iteratefdtd_matrix.zip](https://github.com/UCL/TDMS/files/10286980/iteratefdtd_matrix.zip)
[arc_18.zip](https://github.com/UCL/TDMS/files/10286984/arc_18.zip)
[arc_17.zip](https://github.com/UCL/TDMS/files/10286985/arc_17.zip)
[summary.txt](https://github.com/UCL/TDMS/files/10286986/summary.txt)

**

## ~Blocked by~

* ~#210~
* :heavy_check_mark: #215 

## Potential to conflict with

* the rest of #63

samcunliffe:
## Modifications made to the way the illumination is specified

`iteratefdtd_matrix.m` will create an output file with the following properties, depending on whether a time-domain or pulsed illumination is specified.

There are methods by which the illumination can be specified:

1. When no illumination file is passed to iteratefdtd_matrix as an input parameter, the field can only be pulsed and is specified through the filenames efname and hfname in the input file. There are two options within this case:
   1. Non-empty efname and hfname are specified, resulting in one or more of Isource, Jsource and Ksource being non-empty. tdfield will contain exi and eyi, which will both be empty.
   2. Empty efname and hfname are specified, resulting in all of Isource, Jsource and Ksource being empty. tdfield will contain exi and eyi, which will both be empty.

or

2. When an illumination file containing elements Isource, Jsource and Ksource is specified, created by a call to iteratefdtd_matrix with operation illsetup, illumination will only be pulsed. One or more of Isource, Jsource and Ksource shoudl be non-empty, although it is possible for all to be empty if the user specified empty values for efname and hfname. tdfield will contain exi and eyi, which will both be empty.

or

3. When an illumination file containing fields exi and eyi is specified, a time-domain illumination is specified. tdfield will
contain a copy of exi and eyi. There are two options with regards to pulsed fields:
   1. The user specifies empty efname and hfname, in which case no pulsed field is specified. This is the expected case. Isource, Jsource and Ksource will be empty.
   2. The user specifies non-empty efname and hfname, in which case on or more of Isource, Jsource and Ksource will be non-empty.


[...]

PR #229, opened 2023-02-08
Add system test input file generation to the repository

willGraham01:
## Mostly for my sanity, so that I don't mess anything up when regenerating the tests.

The input data to the system tests is done through a series of scripts and files, none of which are documented nor even tracked in the repository. This makes keeping track of the reference inputs difficult, particularly when new functionality requires us to change the format of the input files themselves (for example, #208 requires us to change the reference inputs but not the corresponding reference _data_).

This PR aims to address this issue.
- The `data_generation` folder is added to the `tdms/tests/system` directory. Each test is a subdirectory within this folder, containing the bare-minimum `.m` and `yaml` files required to regenerate the input data for the corresponding system test.
- A `README.md` is added to the `data_generation` folder with instructions on how to regenerate the input data from the scripts.
- The scripts themselves are standardised so that they run without intermediate modifications to other files within the directory. This presently a problem when separate `illfile`s are used. `iteratefdtd_matrix` requires `efname` and `hfname` to be specified in the illumination `.mat` input, but _not_ in the executable `.mat` input, and previously the `pstd_input_file.m` has to be modified in between generating these two files (otherwise errors would be thrown).

Regenerating the reference _input_ `.mat` files is now a case of re-running the `run_{pstd,fdtd}_bscan.m` scripts within each directory. TODO: Create a script to do this automatically because it is a chore.

## Missing Information!

- `arc_09` is missing it's `pstd_input_file.m` file - this is currently not available from the Zenodo data (nor is available from the old data).
- `arc_10` is missing it's `run_pstd_bscan.m` file - this is currently not available from the Zenodo data (nor is available from the old data).


willGraham01:
Closing in light of conversation w/ @samcunliffe 2023-02-21: a plan has been drawn up for parametrising the MATLAB testing framework and overhauling the Zenodo overloads to reduce cache size.

Issue #230, opened 2023-02-16
Make our project website more public-facing.


samcunliffe:
Following a discussion with @prmunro last week, it'd be good to have a welcoming friendly project website.
Currently, our site (https://github-pages.ucl.ac.uk/TDMS) contains only developer docs and redirects new users to the README.md in the repository.

With doxygen, it's possible to include _multiple_ markdown files which get put in the "Related pages" and included as part of the index...
```bash
# 'mainpage' gets put on the index page
INPUT += README.md
INPUT += CONTRIBUTING.md
INPUT += doc/developers.md
USE_MDFILE_AS_MAINPAGE = README.md
```

Results in:
![Screenshot 2023-02-16 at 15 15 30](https://user-images.githubusercontent.com/1836192/219407404-d4dfc114-df24-462f-a46c-e400dacf3b72.png)

(I.e. contributor guidelines then developer docs (in the order that we'd want because we're lucky with the alphabet).

----

## TODO

- [ ] Make README.md → CONTRIBUTING.md → doc/developers.md flow nicely together (they repeat a bit).
- [ ] Go over README.md and check for clarity.
- [ ] Display department logos and funding agencies?
- [ ] A logo for the web page, or a nice image of light on a sphere, or both.

samcunliffe:
Still some things to iron out. Playing [over on my personal fork of TDMS](https://scnlf.me/TDMS) so as to not break our actual live pages.

willGraham01:
Possibly add a note/request to the end of the readme: (from 28/3/2023 meeting)
If you use TDMS frequently and have example use cases that you don't mind/would like to share, please get in touch at `<foobar>`.

PR #231, opened 2023-02-16
Add to Doxygen pages, and use doxygen as a CI check.

samcunliffe:
This PR is the first part of #230. It makes three changes:

1. Adds README.md and CONTRIBUTING.md to the web pages (they appear in _that_ order in the ToC followed by the developer docs 👍). Screenshot below, and [live dev version over on my personal fork](https://scnlf.me/TDMS).
2. Uses MacOS/homebrew to install and run Doxygen 1.9.6 which is much cooler 😎 than the version we were previously using in ubuntu (1.9.2 👎 ). Amongst other improvements: the cool version correctly renders the `<details>` HTML tags.
3. Splits the Doxygen workflow into two steps: `doxygen` and `publish`. This has a couple of benefits...
    * We can [run the first step (not the second)](https://github.com/samcunliffe/TDMS/actions/runs/4244369222) on all PRs and use it as an additional check that C++ code is documented.
    * The webpages are a build artifact and can be downloaded and browsed and debugged from the action.
    * The [second step can run whenever we want to publish](https://github.com/samcunliffe/TDMS/actions/runs/4244372711). For now, this is on pushes to `main`, but it could be release tags (for example).

In particular for the first step, when we've got everything documented, we can set Doxygen to **fail** when it encounters an undocumented warning. This would then fail the CI.

## Too many words? Here's a screenshot...
Screenshot 2023-02-22 at 21 11 34

## Next planned steps:

* Introduce logos and a citation link.
* Reword README and CONTRIBUTING to avoid duplication of information.
* Turn the Doxygen warnings into GH Actions annotations(?)
* Actually fix Doxygen warnings (there aren't too many of them) and turn Doxygen warnings → build errors.

### DIFF
--- a/.github/workflows/doxygen-gh-pages.yml
+++ b/.github/workflows/doxygen-gh-pages.yml
@@ -36,13 +36,6 @@ jobs:
       run: touch html/.nojekyll
       shell: bash
 
-    - name: Manually copy over assets
-      # Preserve their directory structure to keep README's image references!
-      run: |
-        mkdir -p html/doc/assets
-        cp -R doc/assets/*.png html/doc/assets
-      shell: bash
-
     - name: Upload the website
       uses: actions/upload-artifact@v3
       with:
--- a/.github/workflows/doxygen-gh-pages.yml
+++ b/.github/workflows/doxygen-gh-pages.yml
@@ -1,15 +1,74 @@
-name: Doxygen GitHub Pages Deploy Action
+name: Run Doxygen and publish to GitHub pages
+
+# Generate the documentation on all merges to main, all pull requests, or by
+# manual workflow dispatch. The Doxygen job can be used as a CI check that all
+# functions are documented. The publish job only runs on new commits to the main
+# branch and actually pushes the generated html to the gh-pages branch (which
+# triggers a GitHub pages deployment). When things are a bit more stable, we can
+# only publish the website on release tags.
 on:
   push:
     branches:
       - main
+  pull_request:
   workflow_dispatch:
+
 jobs:
-  deploy:
-    runs-on: ubuntu-latest
+  doxygen:
+    runs-on: macos-latest
     steps:
-      - uses: DenverCoder1/doxygen-github-pages-action@v1.2.0
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          folder: html
-          config_file: doc/Doxyfile
+
+    - name: Checkout repository
+      uses: actions/checkout@v3
+      with:
+        submodules: "true"
+
+    - name: Install Doxygen
+      run: brew install doxygen graphviz
+      shell: bash
+
+    - name: Run Doxygen
+      run: doxygen doc/Doxyfile
+      shell: bash
+
+    # Disabling jekyll ensures pages with underscores work on gh pages.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -34,45 +40,44 @@ $ cmake .. \
...
+find MATLAB, (2) find FFTW or (3) install to the default install prefix.
+
+<details>
+<summary>Mac specific instructions</summary>
+
+To compile on a Mac an x86 compiler with libraries for OpenMP are required,
+which can be installed using [brew](https://brew.sh/) with `brew install llvm`
+then (optionally) set the following cmake arguments
+
+```{sh}
```
willGraham01, line 53:
```suggestion
```bash
```

Minor point, but why `{sh}` here and then `bash` on line 60? Are they subtlety different in how/what they render/highlight?

.github/workflows/doxygen-gh-pages.yml
```diff
@@ -1,15 +1,67 @@
...
+
+    - name: Install Doxygen
+      run: brew install doxygen graphviz
+      shell: bash
+
+    - name: Run Doxygen
+      run: doxygen doc/Doxyfile
+      shell: bash
+
+    # Disabling jekyll ensures pages with underscores work on gh pages.
```
willGraham01, line 34: For my own clarification: you mean pages whose names `contain_underscores` for example?

.github/workflows/doxygen-gh-pages.yml
```diff
@@ -1,15 +1,67 @@
-name: Doxygen GitHub Pages Deploy Action
+name: Run Doxygen and publish to GitHub pages
+
+# Generate the documentation on all merges to main, all pull requests, or by
+# manual workflow dispatch. The Doxygen job can be used as a CI check that all
+# functions are documented. The publish job only runs on new commits to the main
+# branch and actually pushes the generated html to the gh-pages branch (which
+# triggers a GitHub pages deployment). When things are a bit more stable, we can
+# only publish the website on release tags.
```
willGraham01, line 8: :+1:

README.md
```diff
@@ -98,22 +103,27 @@ Options:
 -m:	Minimise output file size by not saving vertex and facet information
```
willGraham01, line 103: I notice that `-c, --cubic-interpolation` are missing from this printout. Although #217 will make this comment redundant anyhow.

willGraham01 (approved):

README.md
samcunliffe: Ah, that's the [doxygen-flavoured markdown way to get syntax highlighting](https://doxygen.nl/manual/markdown.html#md_fenced). Probably this means that on the web page, the line 60 code block is not highlighted.

Although as it's anyway bash, it's probably a bit of a moot point.

.github/workflows/doxygen-gh-pages.yml
samcunliffe: Yes. We don't have any, but we might. And then it'd be a whole nightmare to debug.

README.md
samcunliffe: Looks like GitHub is smart enough to understand both. I'll check when I go over these to check for info duplication.

PR #234, opened 2023-02-21
DRY our .github/workflow

samcunliffe:
It was a bit WET. (Where T=thrice)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,192 @@
...
+#
+# System tests are slow and need input data. To save caching, network, and job
+# time we only run the full system tests on Ubuntu (although this can easily be
+# changed).
+
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
```
willGraham01, line 15:
```suggestion
  pull_request:
    branches: ["main"]
```
Do we not want the tests to run on PRs that target `main`? Or am I just not up to scratch with my GH actions syntax & the `push` field takes care of this?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,192 @@
...
+          sudo apt-get install libfftw3-dev libgomp1 python3
+
+      # -------------------------------------------------------------------------------
+      # Windows
+      - name: Install miniconda for Windows
+        if: ${{ contains(matrix.os, 'windows') }}
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          auto-update-conda: true
+          python-version: 3.9
```
willGraham01, line 72: Just making a note for myself here, but `MATLABEngine` (which the test-regeneration currently uses) is only compatible with `Python >=3.10`. Although given we don't run the tests on Windows, :shrug:

.github/workflows/ci.yml
```diff
@@ -0,0 +1,192 @@
...
+          echo "DOMP_ROOT_MAC=${DOMP_ROOT_MAC}" >> $GITHUB_ENV
+
+      # -------------------------------------------------------------------------------
+      - name: Create build environment
+        run: |
+          cmake -E make_directory ${{ runner.workspace }}/build
+
+      - name: Compile with Werror on Ubuntu and MacOS (not Windows)
+        if: ${{ contains(matrix.os, 'macos') }} || ${{ contains(matrix.os, 'ubuntu')}}
+        run: echo "CXXFLAGS=-Werror" >> $GITHUB_ENV
```
willGraham01, line 102: Does this, plus the existing linting, now resolve #41 too?

willGraham01 (approved):
Mostly a few small comments for my own GitHub actions understanding. Also we might be able to close #41 now?

.github/workflows/ci.yml
samcunliffe: This is: on all pushes to `main`¹, all pull requests, and if a person clicks the button in the actions tab.

 ----
¹ which only happens on PR merge... but we might've allowed direct push to main if we were a smaller team or whatever.

.github/workflows/ci.yml
samcunliffe: Almost, yeah. We should fix whatever's the problem with Windows first though.
I have a Windows machine and VM so I can debug if you like (unless you have a partition?).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,192 @@
...
+  # Only run the slow system tests on Ubuntu.
+  # Cache the input data to save network usage.
+  system-tests:
+    name: System tests ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    needs: build
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest] #, macos-latest]
```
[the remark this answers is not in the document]
willGraham01: I think `fail-fast` being false for the `build` matrix makes sense: if it's an OS-specific problem we get notified of it, and the `system test`s don't start regardless.

That is, I'd rather know how many out of the 6 `builds` worked, rather than only being able to see 1 failed, so the rest stopped.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,192 @@
...
+        uses: actions/cache@v3
+        with:
+          path: ${{ github.workspace }}/tdms/tests/system/data
+          key: test_data
+
+      - name: Run TDMS system tests
+        working-directory: ${{ runner.workspace }}
+        shell: bash
+        run: |
+          export OMP_NUM_THREADS=1
```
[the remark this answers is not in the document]
samcunliffe: Looks like we can have [2 cores for 🐧 🍎 ](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).

Would also be a de-facto solution for #89 .

README.md
```diff
@@ -4,10 +4,8 @@
 and title at the top of README.md when building the project page (the title
 would be duplicated) everything else in README.md is also the project homepage. -->
 
-[![Linux tests](https://github.com/UCL/TDMS/actions/workflows/linux_tests.yml/badge.svg)](https://github.com/UCL/TDMS/actions/workflows/linux_tests.yml)
-[![Windows tests](https://github.com/UCL/TDMS/actions/workflows/windows_tests.yml/badge.svg)](https://github.com/UCL/TDMS/actions/workflows/windows_tests.yml)
+[![Build and test](https://github.com/UCL/TDMS/actions/workflows/ci.yml/badge.svg)](https://github.com/UCL/TDMS/actions/workflows/ci.yml)
```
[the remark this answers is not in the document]
samcunliffe: 😎

.github/workflows/ci.yml
samcunliffe: Your way would mean a PR against a PR would _not_ run the CI. (Perhaps we want it like that?)

.github/workflows/ci.yml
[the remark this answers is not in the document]
giordano: We might even be able to get access to the more beefy machines: https://github.blog/changelog/2023-03-01-github-actions-introducing-the-new-more-powerful-github-hosted-x64-macos-runners/

.github/workflows/ci.yml
willGraham01: Probably want PRs against PRs to run the system tests (although barring the present chain, we're trying to avoid this anyway, so should be a non-issue). I think it's all good as you have it here :+1:

PR #238, opened 2023-02-23
Bugfix GitHub cli new tests

willGraham01:
Getting an error unique to GitHub actions - tests pass locally.

Investigating...
1. ~Is MATLAB actually on the path in the place that I think it is?~ Yes, `which matlab` returns the correct path on the runner, so we _can_ find MATLAB...
2. Does `setup-matlab` actually allow you to _run_ `matlab` from the command-line? Maybe? See update below.

Update:
It looks like MATLAB wants to use display variables (even if you ask it to be suppressed :imp:), but since the `setup-matlab` action presumably setups up MATLAB as `root`, there are no display variables. And even if there were, runners don't have user accounts. For example, trying to run
```bash
matlab -nodisplay -nosplash -nodesktop -r "disp('hello world');exit;"
```
will error on `matlab`'s startup, requesting a display variable. Will try using `matlabengine`, which is the module that allows Python to call MATLAB within `.py` scripts to bypass this issue.

TDLR; you're not allowed to use `matlab -r` or invoke calls to `matlabengine` in Python on GitHub runners. Solutions are:
- Use our own self-hosted runner for the system tests. Will need to have a licensed version of `matlab2022b` and `python{3.10,3.9}` installed for our workflow to operate correctly.
- (Maybe) we can get matlab to read in our config files and run a matlab script that generates the input data. This is a pain because we can no longer link the generation of input data and the system test it corresponds to (we will have to regenerate _all_ of the input data files via the hypothetical `matlab` script, before then running all the system tests via `pytest`.

### DIFF
--- a/.github/workflows/linux_tests.yml
+++ b/.github/workflows/linux_tests.yml
@@ -41,9 +41,10 @@ jobs:
       - name: Set up MATLAB
         uses: matlab-actions/setup-matlab@v1.2.3
 
-      - name: Check MATLAB is on the PATH
+      - name: Check MATLAB is on the PATH and can be invoked
         run: |
           echo $(which matlab)
+          matlab -r -nosplash -nodesktop -nodisplay -r "disp('hello world')"
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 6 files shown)


willGraham01:
Closing as per #239

PR #243, opened 2023-03-01
Add `arc_09` and `arc_10` to the scope of `test_regen.py`

willGraham01:
Continues towards #236 |

### Tests added

`arc_09` and `arc_10` can now be run using `test_regen.py`, and have their input data remade locally.

### `run_bscan` changes:
- `run_bscan` handles the case when the user provides a `.mat` time-domain field as the illumination source (through the `calc_tdfield` flag)
- `run_bscan` now returns the filenames of the input `.mat` files it regenerated, so we can remove all other unnecessary `.mat` files that the input generation process creates.

### Additional changes
- Adds the option to specify a point-source in as the obstacle (`arc_09`)
- Adjusts some argument/input orders to better reflect interdependence between certain variables
- Updates `config_01.yaml`'s example to reflect the new `calc_tdfield` optional flag
- `getsourcecoords` and `calc_field_tdfield` added to repository as necessary functions for the input generation to run


samcunliffe (approved):
Once rebased, all of these ⛓️ guys are easy to review.


... that is to say, the dependent chain of smaller PRs was a good idea.

Issue #240, opened 2023-02-28
Track system test `.mat` input generation.


willGraham01:
We currently do not track the scripts used to generate the input .mat files to the tdms system tests. In light of #217 requiring us to change the input format, we should make the generation of the input data something we track in the repository and incorporate into our system tests, before we make changes and overwrite the old reference inputs.

willGraham01:
Reopening because the "base" of the PR chain was merged in before the final adjustments.

PR #245, opened 2023-03-03
Add `arc_17` and `arc_18` to test coverage

willGraham01:
Moves towards #208 | Continues #240

Adds `arc_17` and `arc_18` to the test coverage.
These tests require an overhaul of how `run_bscan` works, as we no longer simply generate two input `.mat` files per test :( Instead, we now specify the `.mat` inputs we want to generate in the config file, and use `.yaml` syntax to indicate which runs of `tdms` will be using this input file as its... input.

This is recorded in the updates to the `README.md` file. This is also necessary for us to add `arc_17` and `arc_18` to the test coverage.

TODO:
- [ ] Upload to Zenodo the inputs for the new tests and the output data
- [ ] Add `17` and `18` to `test_system.py`, with the old-style config files
- [ ] Update `run_bscan` and repo to actually successfully produce `17` and `18`'s inputs (or at least, those which are appropriate prior to the update to `iteratefdtd_matrix`.


samcunliffe:
codecov/project failure → #279

samcunliffe (approved):

PR #241, opened 2023-02-28
Track generation for `arc_02` and `arc_03`

willGraham01:
Continues #240 |

Adds `arc_02` and `arc_03` to the list of tests which are now covered by the input data regeneration scripts.

Modifications on top of #236:
- `run_bscan` now handles different spatial objects, obstacle radii, and when `illsetup` is required on top of the usual `filesetup`.
- `config_XX.yaml` and `input_file_XX.m` are added to the `data/input_generation` folder appropriately
- ~`generate_input_data.py` has been split into `bscan_arguments.py` and `generation_data.py` containing classes of the same name, to separate functionality.~
- ~Reduces the number of variables and input parameters that were stored more than once across the test classes~ (this is redundant because the classes now no longer exist in the base branch)
- `test_regen.py` will run tests in numerical order when called locally, which is nice.

Force-push changes here :point_right: https://github.com/UCL/TDMS/pull/241#issuecomment-1521771053

PR #257, opened 2023-03-13
`isort` to make `black`-compatible changes

willGraham01:
`isort` uses a default linting style for long includes [that is incompatible with `black`s default](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#isort). We have both in our pre-commit and do not resolve this conflict.

This PR adds a `pyproject.toml` to the repo to implement the recommended fix.

(This conflict has only just come to light because it only occurs when an `import` line is too long and needs to be wrapped. Working on #241 has bought this to light)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[tool.isort]
+profile = "black"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,0 @@
-[tool.isort]
-profile = "black"
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,2 @@
+[settings]
+profile=black


pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[tool.isort]
+profile = "black"
```
samcunliffe, line 2: I would avoid `pyproject.toml` because we're not a python package.

Perhaps the[ `.isort.cfg` option works](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#custom-configuration). It would also be fine by me to just remove `isort` from the pre-commit config file.

samcunliffe (approved):

PR #270, opened 2023-04-26
Add more tests for the argument parsing.

samcunliffe:
An easy train ride and TI planning background task: add some simple tests of arg-parsing functionality.
Not a high priority to review. (Although should be relatively quick.)

Also, and this is not normally good practice, fix a totally unrelated typo 
* #272


tdms/tests/unit/test_argument_parser.cpp
```diff
@@ -0,0 +1,114 @@
+/**
+ * @file test_openandorder.cpp
```
samcunliffe, line 2:
```suggestion
 * @file test_argument_parser.cpp
```

tdms/tests/unit/test_argument_parser.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/** Convert a vector of std::strings to an array of chars */
+const char **vector_to_array(const vector<string> &vec) {
+  const char **arr = new const char *[vec.size()] {};
+  std::transform(vec.begin(), vec.end(), arr,
+                 [](const std::string &s) { return s.c_str(); });
+  return arr;
+}
+// TODO: check with Will that this doesn't shadow the implementation
```
willGraham01, line 39: I am still not super in-the-lingo; but if you're worried that the executable is meant to do this conversion itself, it doesn't.

Issue #191, opened 2022-11-18
DTilde test causes a bus error on MacOS M2


samcunliffe:
Not caught in CI because pathological to MacOS on ARM64. **Not reproducible** on any other OS we've tried so far. Nor Mac M1 (thanks @t-young31).

The [first `SECTION`](https://github.com/UCL/TDMS/blob/3d0fe2c5f1fe6cd1ad917ab7bac4f966472a3553/tdms/tests/unit/array_tests/test_DTilde.cpp#L20-L25) macro in the DTilde test somehow causes a bus error (access to nonexistent memory). So far it seems that [the call to `mxDestroyArray`](https://github.com/UCL/TDMS/blob/3d0fe2c5f1fe6cd1ad917ab7bac4f966472a3553/tdms/tests/unit/array_tests/test_DTilde.cpp#L91) is responsible. 

> **Note**
> Links are at 3d0fe2c5f1fe6cd1ad917ab7bac4f966472a3553 so that they're permanent. Likely this was introduced in earlier commits (#159, #162).

Low priority for now. Especially given that it's likely to be made obsolete by #70, but I'm making a ticket so I don't forget to investigate and check it goes with MATLAB.

## Ideas

 - [ ] Update MATLAB (my Mac has 2022a and the CI runners are on b).
 - [ ] Try `gdb` on linux. `lldb` is not helpful.
 - [ ] Check `matlab_array` is non-empty somehow, intercept.

## Links

Might relate to

* #129

samcunliffe:
So adding debug output it’s the mxDestroyArray that causes it. Trail end of `test_DTilde.cpp`...

```C++
  // tear down
  SPDLOG_DEBUG("About to teardown");
  mxDestroyArray(test_input);
  SPDLOG_DEBUG("Called mxDestroyArray");
}
```
Results in:
```bash
➤ ./tdms_tests "DTilde" 
Filters: DTilde
Randomness seeded to: 1507720952
[2022-11-17 15:26:21.956] [debug] [test_DTilde.cpp:23] num_det_modes() == 0
[2022-11-17 15:26:21.957] [debug] [test_DTilde.cpp:25] !x.has_elements()
[2022-11-17 15:26:21.957] [debug] [test_DTilde.cpp:27] !y.has_elements()
[2022-11-17 15:26:21.957] [debug] [test_DTilde.cpp:30] REQUIRE macro is satisfied
[...]

PR #275, opened 2023-04-27
Update catch2

samcunliffe:
v3.3 has some nice features. Including a `SKIP()` macro. 💙

### DIFF
--- a/tdms/tests/unit/array_tests/test_DTilde.cpp
+++ b/tdms/tests/unit/array_tests/test_DTilde.cpp
@@ -53,6 +53,7 @@ void DTildeTest::test_incorrect_number_of_fields() {
     REQUIRE_THROWS_AS(dt.initialise(matlab_input, 0, 0), runtime_error);
   }
   SECTION("Struct with too few fields") {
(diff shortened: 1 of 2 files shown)


tdms/tests/unit/array_tests/test_DTilde.cpp
```diff
@@ -53,6 +53,7 @@ void DTildeTest::test_incorrect_number_of_fields() {
     REQUIRE_THROWS_AS(dt.initialise(matlab_input, 0, 0), runtime_error);
   }
   SECTION("Struct with too few fields") {
+    SKIP("Causes segmentation violation.");
```
willGraham01, line 56: #191 is still at large. Although I think removing MATLAB will fix this as the weird MATLAB pointers will be cleaned up.

willGraham01 (approved):

tdms/tests/unit/array_tests/test_DTilde.cpp
samcunliffe: Needed to use `SKIP` a lot earlier than I intended 🙃

PR #276, opened 2023-04-28
Typo in README

samcunliffe:
Don't know how this got past codespell. 🤦

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ $ cmake ../tdms \
 $ make install
 ```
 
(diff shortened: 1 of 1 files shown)


willGraham01 (approved):
`installat` : Location at which to install,

Maybe a common enough name that codespell doesn't bat an eyelid?

samcunliffe:
You're right!

```
➤ echo "Today I learned that installat is a word." > test.txt
➤ codespell test.txt 
➤ 
```

PR #277, opened 2023-04-28
Add docstrings to variables

### LINKED ISSUE #219: Physical Meaning of variables
There are a few variables in the codebase that need doc-strings, but we are unsure of the physical object they are describing. These are:

- The `ID` output. It is setup in the main loop [here](https://github.com/UCL/TDMS/blob/1924201bdad21465f6aa05677b4294eed7993832/tdms/src/iterator.cpp#L604-L639)
- The role of [`DTilde`](https://github.com/UCL/TDMS/blob/1924201bdad21465f6aa05677b4294eed7993832/tdms/src/iterator.cpp#L367-L372) and in particular the [`num_det_modes()`](https://github.com/UCL/TDMS/blob/1924201bdad21465f6aa05677b4294eed7993832/tdms/src/iterator.cpp#L606) which seems related to the `ID` output.
- `Pupil` - it's initialised from the frequency vectors [here](https://github.com/UCL/TDMS/blob/1924201bdad21465f6aa05677b4294eed7993832/tdms/src/iterator.cpp#L371) and only appears to be used [here](https://github.com/UCL/TDMS/blob/1924201bdad21465f6aa05677b4294eed7993832/tdms/src/iterator.cpp#L1022-L1026)

willGraham01:
# Context/Description 
Adds docstrings to variables and classes that were currently lacking them, using the information @prmunro provided in #219.

* Fixes #219

## Testing

Only changes are docstrings, and a quick Doxygen build locally confirms the docstrings are found.

### DIFF
--- a/tdms/include/arrays.h
+++ b/tdms/include/arrays.h
@@ -432,17 +432,29 @@ public:
   }
 };
 
+/**
+ * @brief Stores the fibre modes in the Fourier plane of the objective lens.
+ *
+ * The "Tilde" indicates that these quantities are in a Fourier plane relative
+ * to where the optical fibre is actually located, meaning that is has a Fourier
+ * relationship relative to the physical fibre mode(s).
+ */
 class DTilde {
 protected:
-  int n_det_modes = 0;
+  int n_det_modes = 0;//< Number of modes specified
   static void set_component(Tensor3D<std::complex<double>> &tensor,
                             const mxArray *ptr, const std::string &name,
                             int n_rows, int n_cols);
 
 public:
+  /** @brief Fetch the number of modes */
   inline int num_det_modes() const { return n_det_modes; };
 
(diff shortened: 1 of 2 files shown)


samcunliffe (approved):
🍉

tdms/include/arrays.h
```diff
@@ -421,17 +432,29 @@ class Tensor3D {
   }
 };
 
+/**
+ * @brief Stores the fibre modes in the Fourier plane of the objective lens.
+ *
+ * The "Tilde" indicates that these quantities are in a Fourier plane relative
```
samcunliffe, line 438: Sorry, ninja additional suggestion. Did we try out LaTeX & Doxygen?

$\tilde D$ works?

willGraham01: No :(

image

willGraham01:
CodeCov once again hates me too - is it looking at the correct basis?

PR #278, opened 2023-04-28
Handle structs (except with complex data, that's nasty)

willGraham01:
Some progress that I've made on reading structs. They're annoying - `hdf5` treats them as `H5::Group`s which isn't so bad, but then we have to know how to access each member of the group and know a-priori what it is before we attempt to spew its contents into a buffer.

TLDR; `H5Reader` now has a `read_field_from_struct` method. You call it by providing the name of the structure array within the input_file (I have not managed recursive structs yet, and I pray I don't have to), the field within the `struct` that you want, and the datatype that _MATLAB_ saved it as (**NOT** what _you_ or what you'd think `HDF5` would read it as).

Also reading character arrays from `MATLAB` via `HDF5` is fun, you should take a peek at the test for it. Said test (which at least confirms that the reading is working) is found in `tdms/tests/unit/test_hdf5_io.cpp#TEST_CASE("Read from a MATLAB struct")`.

Making this a PR against our dev branch in case @samcunliffe has made any of his own changes locally and this will upset them.


tdms/tests/unit/test_hdf5_io.cpp
```diff
@@ -18,9 +18,9 @@
 
 // tdms
 #include "arrays.h"
-#include "fdtd_grid_initialiser.h"
 #include "unit_test_utils.h"
 
+using namespace std;
```
samcunliffe, line 23: Not a strong comment, but you aren't really benefiting from using the namespace here. I mean: we don't make the code significantly shorter nor save that much whitespace (`std::` is only 5 characters) and you only delete the prefix in a couple of places.

samcunliffe (approved):
Very nice.

tdms/tests/unit/test_hdf5_io.cpp
willGraham01: Yeah sorry, should've been more vigilant when cleaning up my code :sweat_smile: It's my own laziness for moving function names and the like around whilst trying to get MATLAB to play nicely with HDF5

Issue #279, opened 2023-05-03
Coverage reports are inconsistent.


samcunliffe:
**Describe the bug**
The coverage reports collected in `ubuntu-latest` and `macos-latest` differ by 50.0%.

* [MacOS reports 48.2%](https://github.com/UCL/TDMS/actions/runs/4870415033/jobs/8686156342?pr=245#step:14:201)
* [ubuntu reports 24.1%](https://github.com/UCL/TDMS/actions/runs/4870415033/jobs/8686156011?pr=245#step:14:194)

**To Reproduce**
Happens on the build system ([for example here on #245](https://app.codecov.io/gh/UCL/TDMS/pull/245?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=UCL)). A symptom of this is that one of the codecov tests can fail if the 20% report is received first and compared to the 40% report (it looks like a decrease in coverage of 20% which is unacceptable).

**Expected behaviour**
Coverage should be roughly consistent. Some differences are expected due to compiler optimisations and "the deep weeds of machine code". I don't expect such a precise discrepancy though.

**Environment (please complete the following information):**
 - OS: MacOS 12.6.5, and Ubuntu 22.04.2
 - Compiler: AppleClang 14.0.0.14000029 and GNU 11.3.0 
 - MATLAB version (if relevant): R2022a

samcunliffe:
Some things I'm starting to think we should anyway do:

* [Decrease the precision of codecov](https://docs.codecov.com/docs/codecovyml-reference#coverageprecision) and round up.
* Pick `gcc` or `AppleClang` and stick with it.
* [I don't like it, but] also get the coverage from the system tests.

samcunliffe:
Closing as superseded by #305.

PR #284, opened 2023-05-05
Add some configuration to codecov.

samcunliffe:
Adds a [threshold for the patch commit status check](https://docs.codecov.com/docs/commit-status#threshold) (the default is 0). And reduce the precision (quite drastically).

Related to (but not a solution of):
* #279


samcunliffe:
Ah, [I've just read that we could move the file to the .github directory](https://docs.codecov.com/docs/codecov-yaml#frequently-asked-questions) so it doesn't clutter the project (what do you think?)

willGraham01:
> Ah, [I've just read that we could move the file to the .github directory](https://docs.codecov.com/docs/codecov-yaml#frequently-asked-questions) so it doesn't clutter the project (what do you think?)

It's only one line, but I agree. Sadly I can't move files around via GitHub (and my laptop is currently occupied with a rebase) but I'll pre-emptively approve you doing so.

willGraham01 (approved):

Issue #285, opened 2023-05-05
Classes to detach from MATLAB


willGraham01:
Essentially a tracker for our progress towards #70.

# Logging which `tdms` classes need to be detached from MATLAB

## Non-MATLAB classes we might want to flatten
Whilst it won't be necessary (we can just reshape buffers before writing/reading) we might want to rework the classes below to have flat, strided vectors to store their data rather than pointers to pointers.
- [x] `Tensor3D`
- [x] `Matrix`

## Storage classes
Classes that are inherited from by other classes, or are simply used to store or interact with MATLAB objects.
- [x] `MaterialCollection` (#315)
- [x] `CCollectionBase` (#315)
- [x] `DCollectionBase` (#315)
- [ ] `Vector` (this class can almost surely be replaced with `std::vector` now that we can read MATLAB buffers via HDF5).
- [ ] `Dimensions` (this could quite easily be replaced with `H5Dimensions` from #281)
- [ ] `Field`
- [ ] `SplitField`
- [ ] `LoopVariables` (just for the copy of the electric field at the previous iteration)

## Classes that are only read into
Classes that are read from the input file, or initialised by passing `mxArray` pointers.
- [x] `CMaterial` (inherits `MaterialCollection`, `CCollectionBase`) (#315)
- [x] `DMaterial` (inherits `MaterialCollection`, `DCollectionBase`) (#315)
- [x] `DispersiveMultiLayer`
- [ ] `GratingStructure` (inherits `Matrix`)
- [x] `FrequencyExtractVector` (#328)
- [x] `FrequencyVectors`
- [x] `Pupil` (inherits `Matrix`)
- [ ] `DTilde` (inherits `Tensor3D`)
- [ ] `IncidentField` (inherits `Tensor3D`)
- [ ] `fdtdGridInitialiser`
- [x] `InterfaceComponent`
- [x] `Cuboid`
- [ ] `Source` (requires `complex` casting)

## Classes that are written out
Classes that need some (or all) of their data written to the output file.
- [x] `FieldComponentsVectors` (inherits `Vector`, part of `VertexPhasors`) (#328)
- [ ] `Vertices` (inherits `Matrix`, part of `VertexPhasors`)
- [ ] `ElectricField` (inherits `Field`)
- [ ] `MagneticField` (inherits `Field`)
- [ ] `FieldSample`
- [ ] `SurfacePhasors`
- [ ] `VertexPhasors`
- [ ] `IDVariables`

## Classes that are both read and written
[...]

PR #250, opened 2023-03-09
Pair developing HDF5 #181

samcunliffe:
Resolves #181. Relates to #70.

PR #290, opened 2023-05-11
Detach Tensor3D from MATLAB

willGraham01:
# Context/Description 

Detaches `Tensor3D` class from MATLAB. This class is now robust enough to read from either standard `C++` buffers or from `MATLAB` data of the appropriate shape. We also no longer need to use `malloc` for memory management.

The dependant classes still depend on MATLAB for reading/writing from the input file, which can be addressed in another pull request.

* Continues #285 : Resolves `Tensor3D` task and addresses dependant classes.
* Depends on #324 : This will target `main` once this goes in.

#### `Tensor3D` wraps a `std::vector`.
- `operator()` is used to index/reference the data with three indices. 
- We cannot use the `operator[]` since this can only take 1 argument unless we up our `C++` standard to `23`. 
- The `operator[]` can still be invoked if passed an `ijk` `struct` from  `cell_coordinate.h`, but this simply unpacks and passes the information to `operator()`. 
- This causes these changes to be noisy, since we have to change the call methods for all classes that have `Tensor3D` members or parents. 

#### Data is stored as a strided vector
- Row-major format is used since `C` and `hdf5` use this convention. IE Later dimensions have faster-varying indices.
- Using a strided vector means that `zero()`, `frobenius()`, and any other all-element operations can be carried out using `std::vector` methods. 
- The class stores the "3D" dimensions that it is mimicking when they are set.

#### Reading 3D buffers
- `Tensor3D::initialise` allows us to read from `***T` buffers. 
- To comply with `std::vector` ownership laws and avoid memory leaks, data is copied into `Tensor3D`s storage rather than reassigning the underlying pointer.

#### File restructuring
In an attempt to curb the rampant expansion of the `arrays.h` file, `class`es and `struct`s within this file have begun to be broken out into separate files within the `arrays` directory.

## Testing

Unit tests for `Tensor3D`, `DTilde`, and `IncidentField` have been preserved, and continue to pass. `Tensor3D` unit tests have been updated slightly to reflect the new call syntax.

### DIFF
--- a/tdms/include/arrays/tensor3d.h
+++ b/tdms/include/arrays/tensor3d.h
@@ -54,10 +54,10 @@ public:
     initialise(buffer, n_layers, n_cols, n_rows);
   }
 
-  /** 
+  /**
    * @brief Subscript operator for the Tensor, retrieves the (i,j,k)-th element.
    * @note Does not 'call' the tensor, but rather accesses the data.
-   * @example ... 
+   * @example ...
    * @seealso Tensor3D::operator[](const ijk&)
    */
   T &operator()(int i, int j, int k) { return data_[to_global_index(i, j, k)]; }


samcunliffe:
Can this be targetted at `main` yet? Or do we need all of #250 's changes to go in first?

willGraham01:
> Can this be targetted at `main` yet? Or do we need all of #250 's changes to go in first?

Need all of #250 I'm afraid. However this, #303 <- #307, and my current attempts to deal with `XYZVector` are all parallel from there on out.

tdms/tests/unit/array_tests/test_DTilde.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include "array_test_class.h"
 #include "arrays.h"
+#include "arrays/dtilde.h"
```
samcunliffe, line 11: Can we now _not_ `SKIP` the DTilde test? I'd guess: yes.
👇👇 line ~53

tdms/include/arrays.h
```diff
@@ -354,127 +355,6 @@ class Pupil : public Matrix<double> {
   ~Pupil();
 };
 
-template<typename T>
```
samcunliffe, line 357: Nice. Your -741 lines is cool. 😎

tdms/include/arrays/dtilde.h
```diff
@@ -0,0 +1,54 @@
...
+#include <vector>
+
+#include "arrays/tensor3d.h"
+#include "matrix.h"
+
+/**
+ * @brief Stores the fibre modes in the Fourier plane of the objective lens.
+ * @details The "Tilde" indicates that these quantities are in a Fourier plane
+ * relative to where the optical fibre is actually located, meaning that it has
+ * a Fourier relationship relative to the physical fibre mode(s).
```
samcunliffe, line 19:
```suggestion
 * @brief Stores the fibre modes in the Fourier plane of the objective lens.
 * @details The "Tilde" indicates that these quantities are in a Fourier plane
 *          relative to where the optical fibre is actually located, meaning
 *          that it has a Fourier relationship relative to the physical fibre
 *          mode(s).
```

tdms/include/arrays/incident_field.h
```diff
@@ -0,0 +1,22 @@
...
+ * @brief Declaration for the IncidentField object.
+ */
+#pragma once
+
+#include <string>
+
+#include "arrays/tensor3d.h"
+#include "matrix.h"
+
+class IncidentField {
```
samcunliffe, line 12: Could do with a docstring if you have it in your mind.
Although I'm fine if not: that that's a bit of a scope-drift for this PR.

tdms/include/arrays/tensor3d.h
```diff
@@ -0,0 +1,149 @@
...
+  Tensor3D() = default;
+  Tensor3D(int n_layers, int n_cols, int n_rows) {
+    allocate(n_layers, n_cols, n_rows);
+  }
+  Tensor3D(T ***buffer, int n_layers, int n_cols, int n_rows) {
+    initialise(buffer, n_layers, n_cols, n_rows);
+  }
+
+  /** @brief Subscript operator for the Tensor, retrieving the (i,j,k)-th
+   * element. */
```
samcunliffe, line 58: I'd add a bit more detail here: especially since, in many programming languages, the `()` operators would be expected to call a function and `[]` would be expected to access.
```suggestion
  /** 
   * @brief Subscript operator for the Tensor, retrieves the (i,j,k)-th element.
   * @note Does not 'call' the tensor, but rather accesses the data.
   * @example ... 
   * @seealso Tensor3D::operator[](const ijk&)
   */
```
(Please check I have correctly referenced the operator... [I don't think I've ever done that before in doxygen but it looks like this is the way](https://stackoverflow.com/a/52170134/1444054))

tdms/include/arrays/tensor3d.h
```diff
@@ -0,0 +1,149 @@
...
+  T operator[](const ijk &index_3d) const {
+    return data_[to_global_index(index_3d)];
+  }
+
+  /**
+   * @brief Whether the tensor contains any elements.
+   * @details Strictly speaking, checks whether the total_elements() method
+   * returns 0, indicating that this tensor has no size and thus no elements.
+   */
+  bool has_elements() const { return total_elements() != 0; }
```
samcunliffe, line 78:
```suggestion
  bool is_nonzero() const { return total_elements() != 0; }
```

Maybe a better name? (Although you'd need to rename in VSCode).

tdms/tests/unit/field_tests/test_Field_interpolation.cpp
```diff
@@ -27,12 +27,12 @@ using tdms_tests::TOLERANCE;
 We will test the performance of BLi at interpolating a known field to the centre
 of each Yee cell, for both the E- and H- fields. The benchmark for success will
 be superior or equivalent error (Frobenius and 1D-dimension-wise norm) to that
-produced by MATLAB performing the same functionality. Frobenious error is the
+produced by MATLAB performing the same functionality. Frobenius error is the
```
samcunliffe, line 30: 🙃

samcunliffe (approved):
V. minor suggestions: a bit more docs and one thing you might want to rename.

PR #296, opened 2023-05-15
First pass adding basic prerequisites

samcunliffe:
Very basic preliminaries. CMake, Windows SDK, Xcode etc.

~~Opening early in case anyone has time to scrutinise this. Still waiting for some feedback from @prmunro 's postdoc and should also sanity-test these on a Windows machine.~~

Now tested on Windows, and I claim: ready for review.

I'll wait for Ed to email back some comments before merging (if he has a GH account he could even be invited to the review). This is probably good enough to go in around the same time as #320. Then we can patch.


README.md
```diff
@@ -17,55 +17,114 @@ For further details about the method, please refer to the [PDF documentation](ht
...
+
+It might be more straightforward to use the [Windows subsystem for Linux (WSL2)](https://learn.microsoft.com/en-gb/windows/wsl/install), or set up an linux virtual machine.
+
+However, TDMS _can_ be compiled natively on Windows.
+This has been tested Windows 10 with PowerShell.
+
+Assuming you don't already have them, you'll need to download and install:
+
+* [MATLAB](https://www.mathworks.com/products/matlab.html),
+* [Visual Studio](https://visualstudio.microsoft.com/vs/community/) and be sure to select the C++ kit,
```
willGraham01, line 71:
```suggestion
* [Visual Studio](https://visualstudio.microsoft.com/vs/community/) and **be sure to select the C++ kit**,
```

I feel like this should carry this emphasis since if you're like me (or the students that attend Carpentries workshops), you see a list of prerequisites and you don't look past the link you're given to the installer. Bold text helps draw the eye to "there's more to this step than you think".

README.md
```diff
@@ -17,55 +17,114 @@ For further details about the method, please refer to the [PDF documentation](ht
...
 ```
+
+Which will help Windows locate `.dll` files later.
+For all following instructions, you'll have to substitute our mentions of `tdms` with `tdms.exe` and `$` is used to denote a command prompt which, in PowerShell would look like `PS>`
+
 </details>
 
+------
+
+You'll need to download and install [MATLAB](https://www.mathworks.com/products/matlab.html), and take note where the headers are installed.
```
willGraham01, line 104: If we're saying this, we might as well move it into the install list above (like when we say you need VSCode _and_ then do this other thing while you install it).

```md
Assuming you don't already have them, you'll need to download and install:

* [MATLAB](https://www.mathworks.com/products/matlab.html), and **take note where the headers are installed**
```

willGraham01 (approved):
Approving, but obviously still pending feedback from Peter's postdoc/PhDs

README.md
samcunliffe: Ah but this one is common to all three `<details>` sections. As in: do all of the things, and then get MATLAB.

samcunliffe: Perfectly happy to put it up into the sections though. We also say something similar _before_ all of the `<details>` blocks.

README.md
```diff
@@ -102,7 +161,7 @@ In a terminal run
 $ which tdms
 ```
 
-Copy the full path (something like `/usr/local/bin/tdms`) into [`run_pstd_bscan.m`](https://github.com/UCL/TDMS/blob/main/examples/arc_01/run_pstd_bscan.m), replacing the `'tdms'` text in the calls to the `system()` function.
+Copy the full path (something like `/usr/local/bin/tdms`) into [`run_pstd_bscan.m`](https://github.com/UCL/TDMS/blob/main/examples/arc_01/run_pstd_bscan.m), replacing the ` 'tdms' ` text in the calls to the ``system()`` function.
```
samcunliffe, line 164: Something weird happened with your (good) suggestion on the [doxygen rendering](https://github-pages.ucl.ac.uk/TDMS/index.html#autotoc_md2). Perhaps this fixes...?

README.md
```diff
@@ -17,55 +17,114 @@ For further details about the method, please refer to the [PDF documentation](ht
...
+# -DCMAKE_INSTALL_PREFIX=$HOME/.local/
+$ cmake --build . --target install --config Release
+```
+
+If CMake cannot find MATLAB, FFTW, or install to the default installation prefix, uncomment the relevant line(s) and modify the path(s) accordingly.
+
 You can check that `tdms` was installed correctly and is in your `PATH` by running:
 ```{sh}
-$ tdms --help
+$ tdms -h
```
samcunliffe, line 127: `--help` actually isn't recognised (but perhaps it should be).

README.md
```diff
@@ -17,55 +17,114 @@ For further details about the method, please refer to the [PDF documentation](ht
...
+```
+
+You'll need to ensure the paths to FFTW and MATLAB (the locations of `fftw3.dll` and `libmex.dll` respectively) are in the `env:Path`.
+
+These can be found, e.g. by
+```{pwsh}
+PS> conda list fftw # assuming you installed via conda
+PS> which.exe MATLAB
+```
+Which should return something like `C:\Program Files (x86)\MATLAB\R20XXx\bin\matlab` and maybe `C:\ProgramData\envs\base\bin`.
```
samcunliffe, line 88:
```suggestion
Which should return something like `C:\Program Files (x86)\MATLAB\R20XXx\bin\` for MATLAB, and `C:\ProgramData\anaconda\envs\base\Library\lib` or `C:\Users\<your username>\anaconda\Library\lib\cmake\fftw3`.
```

README.md
samcunliffe, line 127:
```suggestion
$ tdms --help
```

README.md
```diff
@@ -17,52 +17,111 @@ For further details about the method, please refer to the [PDF documentation](ht
...
 $ arch -x86_64 zsh
 $ arch -x86_64 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
-$ arch -x86_64 /usr/local/bin/brew install llvm hdf5
+$ arch -x86_64 /usr/local/bin/brew install cmake fftw llvm
 ```
+
 </details>
 
+<details>
+<summary><img src="https://github.com/EgoistDeveloper/operating-system-logos/blob/master/src/24x24/WIN.png"/> Windows prerequisite setup</summary>
```
samcunliffe, line 58: Tagging @edggjames FYI...  👇

README.md
```diff
@@ -17,52 +17,136 @@ For further details about the method, please refer to the [PDF documentation](ht
...
+Support for Windows is quite new and experimental (please [report](https://github.com/UCL/TDMS/issues/new/choose) any issues you encounter!).
+
+It might be more straightforward to use the [Windows subsystem for Linux (WSL2)](https://learn.microsoft.com/en-gb/windows/wsl/install), or set up an linux virtual machine.
+
+However, TDMS _can_ be compiled natively on Windows.
+This has been tested Windows 10 and 11, with PowerShell.
+
+Assuming you don't already have them, you'll need to download and install:
+
+<!-- * HDF5 -->
```
willGraham01, line 74:
```suggestion
<!-- * TODO: When HDF5 comes in -->
```

willGraham01 (commented):
Won't pass explicit approval to prevent the merge queue drinking this up.

But looks fine, weird that the markdown section tags are rendering though. Ping me when things are addressed & I'll provide approval

README.md
samcunliffe: Huh. I had an even better comment with a link to the page... lost.

doc/developers.md
```diff
@@ -107,7 +108,23 @@ ls .git/hooks
 
 ## Compiling and debugging {#compiling}
```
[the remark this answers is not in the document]
samcunliffe: Yeah we use them. And yeah: doxygen hides them and uses them as anchors, but GH doesn't hide them.

So the GH page looks a bit weird.

samcunliffe:
> I'll provide approval and get this into the queue.

Too much #DevelopmentVelocity. I'll send a followup PR.

PR #301, opened 2023-05-17
Let's see if `gcovr` is more illustrative reporting.

samcunliffe:
# Context/Description

Some strange results with `lcov` to summarise the coverage reports. Seems to be unrelated to `lcov` but during debugging we discovered `gcovr` which is a little more readable and needs fewer command-line options to do what we want.

* Relates to #279


samcunliffe:
Closing as not working as intended (despite the passing build).

PR #303, opened 2023-05-18
Detach the `EHVec` class from `Matrix`

willGraham01:
# Description 

The `EHVec` class is currently a public `Matrix<fftw_complex>`. It is desirable to convert `Matrix<>` to used strided vector storage and remove any MATLAB dependencies.

The former doesn't work for `EHVec` since the `fftw_complex` type is a double[2], which throws an error when used in templated `std::vector<T>` functions. Additionally, `EHVec` is used in such a way that `fftw` needs a continuous buffer to perform the forward/backward transform, which a strided vector does not guarantee.

As such, it is necessary to provide `EHVec` with it's own (MATLAB-free) class. This enables `Matrix` and its other derivatives to be freed of MATLAB influence.

* Concerns #285, resolves the `EHVec` task.
* Depends on #324, targets the expected state of `main`.

## Testing

- Existing unit tests for `EHVec` continue to pass after the refactor. 
- Removed the `EHVecTest` class since it is no longer a member of the `arrays.h` file, and it only had a single unit test in the first place (so a class was overkill).

### DIFF
--- a/tdms/include/arrays.h
+++ b/tdms/include/arrays.h
@@ -524,14 +524,6 @@ public:
 class CCoefficientMatrix : public Matrix<double> {};
 
 /**
- * Temporary storage 'vector'
- */
-class EHVec : public Matrix<fftw_complex> {
-public:
-  ~EHVec();
-};
-
-/**
  * Container for storing snapshots of the full-field
  */
 class FullFieldSnapshot {
--- a/tdms/include/arrays/eh_vec.h
+++ b/tdms/include/arrays/eh_vec.h
@@ -11,37 +11,14 @@ protected:
 
   fftw_complex **data_ = nullptr;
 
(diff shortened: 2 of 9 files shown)


samcunliffe (approved):
Totally fine with this... but I'm a bit sad that I can't have a `Matrix<std::complex>`.

What was the error?

willGraham01:
> Totally fine with this... but I'm a bit sad that I can't have a `Matrix<std::complex>`.
> 
> What was the error?

`Matrix<std::complex>` doesn't complain - it's `Matrix<fftw_complex>` which does. This is because `fftw_complex` is a typedef of `double[2]`, and then some obscure error that I think boils down to the fact that an `fftw_complex` is a `double*` so can't be used in the template.

PR #309, opened 2023-05-19
Remove legacy FDTD bootstrapper variables and functions

### LINKED ISSUE #214: `iwave_` bootstrapping variables are unneccessary
The variables used in the call to `extractPhasorsPlane` in `iterator.cpp` are never referenced again in the code, are not outputs that are written to the file, and are not used by any other structures.

At a functional level, this means that they currently serve no purpose in the codebase and we should be able to remove them without issue. However it looks like this is a remnant of a feature that was partially implemented, and then abandoned. Most likely functionality to allow the user to extract phasors on a pre-defined plane - @prmunro can you remember anything about this?

willGraham01:
Closes #214

Removes all references and classes associated with the legacy `FDTDBootstrapper` class, as per the issue described above.

TDMS will run a little bit faster :smile:


samcunliffe (approved):

samcunliffe:
Probs want to remove that header dependence too.

- [ ] Test file presumably has a `#include "simulation_manager/fdtd_bootstrapper.h"`

PR #304, opened 2023-05-19
Add dependence on `libhdf5` and basic implementation class structure.

samcunliffe:
A cherry-pick of the first few commits of #250 for factorised review.

First part of (does not solve)
* #181 

Supercedes
* #193
*  #250

Issue #291, opened 2023-05-11
Add a mention of MATLAB output file version 7.3


samcunliffe:
Because MATLAB doesn't seem to use semver correctly, MATLAB file version 7.3 is the magic version for which everything works, and 7.0 is no good.

This should be explained somewhere in the user-facing docs. Possibly with screenshots of MATLAB for how to change the default.

samcunliffe:
Fixed by #310

PR #310, opened 2023-05-20
Test-driven reading and writing basic datatypes to HDF5.

samcunliffe:
A cherry-pick of the next few commits of #250 for factorised review. This goes up to (not including) the merge commit from #278 so with the next chunk we'll arrive at changes that have already been reviewed as nested PRs to the pairdev branch.

Second part of (does not solve)
* #181 

Continues from
* #304

Supercedes
* #193
*  #250 

## Actual code notes.

Adds a small example test data file to the C++ source code dir, and some infrastructure in CMake to find the path to that file. Adds tests.


tdms/src/hdf5_io.cpp
```diff
@@ -50,7 +70,10 @@ void HDF5Base::ls() const {
   return;
 }
 
+// IJKDimensions HDF5Base::shape_of(const std::string &dataname) const {
+// return to_ijk(dimensions);
```
samcunliffe, line 74:
```suggestion
```

tdms/tests/unit/test_hdf5_io.cpp
```diff
@@ -5,35 +5,47 @@
...
 #include <catch2/catch_test_macros.hpp>
 #include <spdlog/spdlog.h>
 
 // tdms
+#include "arrays.h"
+#include "fdtd_grid_initialiser.h"
 #include "unit_test_utils.h"
 
-using tdms_tests::create_tmp_dir;// unit_test_utils.h
+using tdms_tests::create_tmp_dir;
```
samcunliffe, line 24:
```suggestion
using tdms_tests::create_tmp_dir; // unit_test_utils.h
```

tdms/tests/unit/test_hdf5_io.cpp
```diff
@@ -5,35 +5,47 @@
...
+
+TEST_CASE("Wrong datatype passed to ijk.") {
+  auto tmp = create_tmp_dir();
+  HDF5Writer r(tmp.string() + "/why.h5");
+  // r.data_dump("x");
```
samcunliffe, line 29:
```suggestion
```

tdms/tests/unit/test_hdf5_io.cpp
```diff
@@ -5,35 +5,47 @@
...
-  // setup - temporary directory
+
+  // test-case wide setup - temporary directory
   auto tmp = create_tmp_dir();
 
   SECTION("Check file creation.") {
     HDF5Writer f(tmp.string() + "/test_file_constructor.h5");
     CHECK(f.is_ok());
-  }// destructor called as we leave scope
+  }
```
samcunliffe, line 40:
```suggestion
  }// Destructor called as we leave scope
```
IMHO it's a useful comment to keep.

tdms/src/hdf5_io.cpp
```diff
@@ -1,11 +1,31 @@
...
+ * @return ijk The dimensions in a struct.
+ */
+ijk to_ijk(const std::vector<hsize_t> dimensions) {
+  unsigned int rank = dimensions.size();
+  ijk out;
+  if (rank > 0) out.i = (int) dimensions[0];
+  if (rank > 1) out.j = (int) dimensions[1];
+  if (rank > 2) out.k = (int) dimensions[2];
+  if (rank > 3) spdlog::warn("Rank > 3");
+  return out;
```
willGraham01, line 26: This function gets removed later down the line, but for the sake of resolving conflicts and possible tests, it can stay for now.

tdms/src/hdf5_io.cpp
```diff
@@ -60,10 +83,12 @@ std::vector<hsize_t> HDF5Base::shape_of(const std::string &dataname) const {
...
 }
 
-
-bool HDF5Base::is_ok() const { return true; }
+bool HDF5Base::is_ok() const {
+  // TODO: check for file health might be unnessicary given we've constructed
+  // the object.
+  return file_->isHdf5(filename_);
+  // return file_->isAccessible(filename_) && file_->isHdf5(filename_);
+}
```
willGraham01, line 94: He's back, but I'm fairly sure he's going to disappear again in later commits.

willGraham01 (approved):

tdms/tests/unit/test_hdf5_io.cpp
```diff
@@ -44,29 +56,151 @@ TEST_CASE("Test file I/O construction/destruction.") {
...
+  SECTION("Check can't read nonexistent data.") {
+    {
+      HDF5Writer fw(tmp.string() + "/this_file_does_exist_but_is_empty.h5");
+      CHECK(fw.is_ok());
+
+    }// Destructor called as we leave scope.
+
+    double data[1];
+    HDF5Reader fr(tmp.string() + "/this_file_does_exist_but_is_empty.h5");
+    CHECK_THROWS(fr.read("nonexistantdata", data));
```
samcunliffe, line 68: Not the end of the world, but we're actually not sure which exception is being thrown here. Presumably there'd also be a HDF5 exception if `data[1]` was too small or the wrong type or whatever.

tdms/tests/unit/test_hdf5_io.cpp
```diff
@@ -44,29 +56,151 @@ TEST_CASE("Test file I/O construction/destruction.") {
...
+  }
+
   // Normal operation: we should be able to create a file and write to it, then
   // read from it.
   SECTION("Check write then read.") {
-    // create a file
     {
```
samcunliffe, line 74:
```suggestion
    // Create a file and write some data.
    {
```

tdms/tests/unit/test_hdf5_io.cpp
```diff
@@ -44,29 +56,151 @@ TEST_CASE("Test file I/O construction/destruction.") {
...
+    {
+      HDF5Writer f2(tmp.string() + "/test_file_wor.h5");
+      hsize_t dimensions[1] = {1};
+      double writeme = 54321.;
+      f2.write("testdata2", &writeme, 1, dimensions);
+      SPDLOG_DEBUG("Written second data");
+
+      CHECK(f2.is_ok());
+
+    }// destructor called as we leave scope
```
samcunliffe, line 115:
```suggestion
    }// Destructor called as we leave scope.
```

tdms/src/hdf5_io.cpp
samcunliffe: Not in the header though afai can see.

tdms/tests/unit/test_hdf5_io.cpp
```diff
@@ -44,29 +56,151 @@ TEST_CASE("Test file I/O construction/destruction.") {
...
+#endif
+
+TEST_CASE("Read from example input") {
+  spdlog::info("I'm going to try and find: {}", TESTDATA);
+  HDF5Reader MATFile(TESTDATA);
+
+  SECTION("Read FDTD grid") {
+    fdtdGridInitialiser gridinitialiser;
+    MATFile.read(gridinitialiser);
+    // CHECK();
```
samcunliffe, line 204: Why didn't we `CHECK` here, do you remember?

tdms/include/hdf5_io.h
```diff
@@ -8,12 +8,17 @@
...
+#include <stdexcept>
 #include <string>
 #include <vector>
 
 #include <H5Cpp.h>
 #include <spdlog/spdlog.h>
 
+#include "arrays.h"
+#include "cell_coordinate.h"
+#include "fdtd_grid_initialiser.h"
```
samcunliffe, line 20: Unused? Because we don't return `IJKDimensions` from any function here.

samcunliffe (commented):
It would be nice to make this a coherent change before the next cherry-picks. I'll do some tidy up.

tdms/tests/unit/test_hdf5_io.cpp
willGraham01: HDF5 (un)helpfully wraps all it's errors into HDF5Errors which aren't particularly readable. In this case I think it's their equivalent of a "file not found" error but it's not actually called that :roll_eyes:

tdms/src/hdf5_io.cpp
samcunliffe: Correction: it _is_ in the header. And we do actually use it.

tdms/include/hdf5_io.h
```diff
@@ -66,11 +71,11 @@ class HDF5Base {
    * @brief Return shape/dimensionality information about the array data stored
    * with `name`.
    * @param dataname The name of the data table.
-   * @return std::vector<hsize_t> The dimensions of the data.
+   * @return IJKDimensions The dimensions of the data.
```
samcunliffe, line 74:
```suggestion
   * @return std::vector<hsize_t> The dimensions of the data.
```

tdms/tests/unit/test_hdf5_io.cpp
samcunliffe: Ah, I just mean that we're expecting some kind of specific data exception or something. H5 seems to have a [pretty nice set of exceptions](https://docs.hdfgroup.org/archive/support/HDF5/doc1.8/cpplus_RM/class_h5_1_1_exception.html).

"File not found" is probably a `H5::FileIException`. This non-existant data is _probably_ something like a `H5::DataSetException`. So we should specify the thing we expect to be thrown is thrown. Not just any general exception.

samcunliffe: Ugh. They're both `H5::FileIException`s. Which I guess makes sense.... trying to read something from *a file* that isn't there.

willGraham01: Catch2 has a [`{REQUIRE,CHECK}_THROWS_WITH`](https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md#exceptions) if you want to be super-specific and check the exception message to distinguish the cases?

samcunliffe: I went for `CHECK_THROWS_AS` but not `WITH` (because the exception string is quite long).

tdms/tests/unit/test_hdf5_io.cpp
willGraham01: Did we just forget to write the checks that the data has been read in correctly?

I presume this is meant to be something like
```c++
for(int i = 0; i < n_rows; i++) {
  for(int j = 0; j < n_cols; j++) {
    for(int k = 0; k < n_layers; k++) {
      CHECK(gridinitialiser.materials[i][j][k] == whatever_is_in_the_test_data);
    }
  }
}
```
?

willGraham01:
> Oh wait.
> 
> > THIS SHOULD INCLUDE A NOTE ABOUT THE MAGIC `-v7.3`! #291
> 
> We need this (in this PR? or can it come separately?)

We maybe want to put this as a note in this PR - a short one-liner in the README should suffice (since we need users, not just developers, to be aware of this).

If we want something more detailed, then an addendum-commit to [281](https://github.com/UCL/TDMS/pull/250/commits/eac765d1e1e1c37b1d63049e9649dc753a7a7542) is probably the place to put it, since this is where we get the CI to produce the HDF5 test data via scripts. Maybe a note in doc/developers.md in the unit testing section?

samcunliffe:
> At this point, it's just a one-liner:

 Wondering about adding a screenshot of the preferences

Screenshot 2023-05-22 at 11 15 42

tdms/tests/unit/test_hdf5_io.cpp
samcunliffe: Ah right, it's because #278 actually implements the group stuff (and deletes/abandons the `fdtdGridInitialiser` test file). For now, we were just checking that we could find the file.

I'm opting to throw a not implemented exception and explicitly skip this test for now.

willGraham01:
> Wondering about adding a screenshot of the preferences

Also not a bad idea, although I'd argue we should point them to the Mathworks site that tells them how to do this (if it can be found...). 

Adding MATLAB GUI instructions to our site seems a bit odd if they're available somewhere else and detracts from `TDMS` itself

tdms/src/hdf5_io.cpp
```diff
@@ -51,6 +53,7 @@ void HDF5Base::ls() const {
 }
 
 std::vector<hsize_t> HDF5Base::shape_of(const std::string &dataname) const {
+  SPDLOG_DEBUG("shape_of");
```
willGraham01, line 56: I feel like we've touched on this before, but do we have a consistent style between the `SPDLOG_{macro}`s and `spdlog::{functions}`? 

Not a major thing, but it's probably good to stick to one or the other. And at a glance, the source code uses the functions so perhaps we go with them?

willGraham01 (approved):

tdms/src/hdf5_io.cpp
samcunliffe: Ah yes!

I like functions in the source, and macros are fine (and arguably more helpful) in the tests.

PR #311, opened 2023-05-22
Handle structs (except with complex data, that's nasty) 

samcunliffe:
This PR is a single `git cherry-pick` of the merge commit of #278 but now targeting `main`.

There are a few ways we could tackle this (including rebasing #250). I think one fresh PR per merge commit is the sanest and my favourite way. There are 4 PR merge commits to the pairdev branch so I'll open 4 PRs which are `cherry-picked` clones of your originals.

Part of 

* #181 

Continues from
1. #304 
2. #310 

Supercedes
* #193
* #250 

## Actual code notes and change description

... are in 
* #278 

@willGraham01 please double check the cherry-pick worked and make a final final pass review.


tdms/tests/unit/test_hdf5_io.cpp
```diff
@@ -188,21 +187,93 @@ TEST_CASE("Test w/r TDMS objects") {
   std::filesystem::remove_all(tmp);
```
samcunliffe, line 187: I did take the liberty of keeping the namespace.

willGraham01 (approved):
Technically this is me reviewing my own PR... but I won't tell if you don't.

PR #315, opened 2023-05-23
Detach `XYZVector` from MATLAB

willGraham01:
# Detaches `XYZVector` class from MATLAB

The XYZVector class was previously a collection of three `double *` pointers, which were cast to `mxGetPr` or similar.

This PR allows `HDF5Reader.read()` to read an XYZVector, rather than us needing to rely on the MATLAB API. 
- Continues #285 | Resolves the `MaterialCollection`, `{C,D}CollectionBase`, `{C,D}Material` tasks.
- Depends on #324 

It also detaches the following classes from MATLAB, which depended on `XYZVector`:
- `dispersive_multilayer` : This was already detached, but now we can safely allocate the members and it has been broken out of `arrays.h`.

## `material_collections.h`
This file reworks the old `MaterialCollection`, `CCollectionBase`, `CCollection`, `CMaterial`, `DCollectionBase`, `DCollection`, `DMaterial` classes.
There are now just 4 structs to hold the data that these classes previously contained, `CMaterial`, `CCollection`, `DMaterial` and `DCollection`. The two "`D`" objects are actually identical save for the group in the input file that they read from, so are just typedefs of a slightly more general `DBase` class so that we avoid repeating definitions for the `HDF5Reader.read()` method.

Speaking of which, `HDF5Reader.read()` can now read in these 4 new classes.

## Testing

Existing tests have been preserved, where possible. Since the majority of the new classes in `material_collections.h` are now simply `structs` that reply on `std::vector` functionality, their unit tests are no longer present.

In their place are unit tests for the new overloads of the `HDF5Reader.read()` method, for reading each of the new classes and reworked `XYZVector` class in.


tdms/include/arrays/dispersive_multilayer.h
```diff
@@ -0,0 +1,34 @@
...
+ */
+#pragma once
+
+#include <vector>
+
+#include "arrays/xyz_vector.h"
+
+/**
+ * @brief Stores dispersive constants that describe the properties of the
+ * multilayered structure being simulated.
```
samcunliffe, line 15:
```suggestion
 * @brief Stores dispersive constants that describe the properties of the
 *        multilayered structure being simulated.
```

tdms/include/arrays/xyz_vector.h
```diff
@@ -0,0 +1,64 @@
...
+        break;
+      case 'y':
+        return y;
+        break;
+      case 'z':
+        return z;
+        break;
+      default:
+        throw std::runtime_error("Component not recognised");
+        break;
```
samcunliffe, line 40:
```suggestion
      case 'x':
        return x;
      case 'y':
        return y;
      case 'z':
        return z;
      default:
        throw std::runtime_error("Component not recognised");
        break;
```
On older compilers, I claim we'd even get `Wunreachable-code` warnings.

tdms/include/hdf5_io/hdf5_base.h
```diff
@@ -1,5 +1,5 @@
 /**
- * @file hdf5_io.h
+ * @file hdf5_base.h
```
samcunliffe, line 2: 👀

tdms/src/simulation_manager/objects_from_infile.cpp
```diff
@@ -80,16 +82,17 @@ IndependentObjectsFromInfile::IndependentObjectsFromInfile(
...
+  INPUT_FILE.read("conductive_aux", "rho_", rho_cond);
+  // rho_cond.x =
+  //         mxGetPr(ptr_to_vector_in(matrices_from_input_file["conductive_aux"],
+  //                                  "rho_x", "conductive_aux"));
+  // rho_cond.y =
+  //         mxGetPr(ptr_to_vector_in(matrices_from_input_file["conductive_aux"],
+  //                                  "rho_y", "conductive_aux"));
+  // rho_cond.z =
+  //         mxGetPr(ptr_to_vector_in(matrices_from_input_file["conductive_aux"],
+  //                                  "rho_z", "conductive_aux"));
```
samcunliffe, line 95:
```suggestion
```

tdms/src/simulation_parameters.cpp
```diff
@@ -70,7 +73,9 @@ void SimulationParameters::set_Np(FrequencyExtractVector &f_ex_vec) {
 void SimulationParameters::unpack_from_input_matrices(
         InputMatrices in_matrices) {
   // determine if we have a dispersive medium or multilayer
-  CCollection C(in_matrices["C"]);
+  CCollection C;
+  HDF5Reader INPUT_FILE(in_matrices.input_filename);
+  INPUT_FILE.read(C);
```
samcunliffe, line 78: Why are you shouting? `INPUT_FILE` is not global, right?

samcunliffe (approved):
I think I can reduce your LoC a little.
Otherwise: very nice work!

tdms/src/simulation_parameters.cpp
willGraham01: No, but this is currently an ugly bit of code - creating multiple instances of `HDF5Reader` to setup `SimulationManager` in each of `SimulationParameters`, `OutputMatrices` and `InputMatrices`.

Ideally `INPUT_FILE` would be passed from SimulationManager`.

PR #320, opened 2023-05-24
Fix example script

willGraham01:
# Context

Updates `examples/arc_01` so that it runs and produces an output. The example had been neglected with the update to `iteratefdtd_matrix`.

* Fixes #316 

I wanted to get MATLAB to also run the demonstration script, but it is not as simple as passing the script to the `matlab-actions/run-command@v1` action because:
- We don't install tdms in the current CI version
- MATLAB's `system` command cannot locate the tdms executable without it being in the system path.

Issue #292, opened 2023-05-11
Pedagogical instructions (+ specifically help for Windows users)


samcunliffe:
Add some more basic prerequisite installation instructions.

Can use the CI as a template. For Windows, it might make the most sense to recommend WSL. If not then SDK for C++ and CMake, and... conda is the easiest way to get FFTW.

Probably keep this issue paused until we get feedback from @prmunro's postdoc.

## Subtasks related to this

- [x] #291 
- [ ] #313

samcunliffe:
Make may not exist so should change instructions to `cmake --build . `
FFTW via conda goes to the homedir (even if installing globally!?)

- [ ] change instructions to `cmake --build . `
- [ ] hint for how to find fftw just installed via conda

samcunliffe:
Can reproduce En's issue with missing dlls.

PR #323, opened 2023-05-24
Add troubleshooting instructions to README.md

samcunliffe:
Related to the discussion in
 * #320 

I propose to add some "Troubleshooting" instructions to README.

## Question

Should the MATLAB file format v7.3 instructions also go _inside_ the troubleshooting `<details>` tags? I.e. Hidden by default.

Pros:
* Makes the README shorter and clearer(?)
* Shorter is gooder.
* Not everyone will want to save files in MATLAB.
* MATLAB file format issues are, strictly speaking, not our fault or our business.

Cons:
* The important help could get overlooked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ command not found: tdms
 ```
 
 Assuming `tdms --help` works in a new terminal, try launching MATLAB _from_ that terminal.
- 
+
 ```{sh}
 $ tdms --help
 $ /Applications/MATLAB_<version>.app/bin/matlab
@@ -105,7 +105,7 @@ $ which tdms
 Copy the full path (something like `/usr/local/bin/tdms`) into [`run_pstd_bscan.m`](https://github.com/UCL/TDMS/blob/main/examples/arc_01/run_pstd_bscan.m).
 
 </details>
- 
+
 ### MATLAB file version
 
 In order to be readable by TDMS, files need to be saved in .mat (MATLAB file) version 7.3 or newer.
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ To compile TDMS on a Mac, you will need an x86 compiler with libraries for OpenM
 You can install these using [Homebrew](https://brew.sh) with the command:
 
 ```{sh}
-brew install llvm
+$ brew install llvm
 ```
 
 After installing with Homebrew, you may need to set the following CMake arguments:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,6 +77,35 @@ For beginners, we recommend starting with the demonstration MATLAB script, which
 Move into this directory, launch MATLAB, and run the MATLAB script [`run_pstd_bscan.m`](https://github.com/UCL/TDMS/blob/main/examples/arc_01/run_pstd_bscan.m).
 This script will generate the input to TDMS, run TDMS, and display sample output.
 
+<details>
+<summary>Troubleshooting</summary>
+
+We've seen that launching MATLAB on MacOS via the launcher (cmd + space) may not preserve the system `PATH`.
```
willGraham01, line 83:
```suggestion
We've seen that launching MATLAB on MacOS via the launcher (`cmd` + `space`) may not preserve the system `PATH`.
```

README.md
```diff
@@ -76,6 +77,35 @@ For beginners, we recommend starting with the demonstration MATLAB script, which
...
+The MATLAB example scripts should then find `tdms`.
+If you still have problems, you can try hard-coding the full path to `tdms` into the MATLAB script.
+
+In a terminal run
+
+```{sh}
+$ which tdms
+```
+
+Copy the full path (something like `/usr/local/bin/tdms`) into [`run_pstd_bscan.m`](https://github.com/UCL/TDMS/blob/main/examples/arc_01/run_pstd_bscan.m).
```
willGraham01, line 105:
```suggestion
Copy the full path (something like `/usr/local/bin/tdms`) into [`run_pstd_bscan.m`](https://github.com/UCL/TDMS/blob/main/examples/arc_01/run_pstd_bscan.m), replacing the `'tdms'` text in the calls to the `system()` function.

willGraham01 (approved):
To answer the question posed in the PR: it's probably worth mentioning the save-file format thing somewhere visible, then referring readers to the troubleshooting section from there (since this is quite a fundamental problem that blocks use of TDMS, but is simultaneously not our fault).

Perhaps more generally, the README is getting _quite_ long now. Maybe we should contemplate the idea of `doc/troubleshooting.md`? (and any other sensible breakouts while we're at it)

samcunliffe:
> To answer the question posed in the PR: it's probably worth mentioning the save-file format thing somewhere visible, then referring readers to the troubleshooting section from there (since this is quite a fundamental problem that blocks use of TDMS, but is simultaneously not our fault).

Yeah, OK. Let's leave as is for now then.

 
> Perhaps more generally, the README is getting _quite_ long now. Maybe we should contemplate the idea of `doc/troubleshooting.md`? (and any other sensible breakouts while we're at it)

Well #292 / #296 will make it worse 😸

I'm open to troubleshooting.md but perhaps a longish readme is unavoidable. Maybe better to have everything in one place rather than referencing back and forth to a separate help page.

Let's merge this as is for now.

PR #328, opened 2023-06-02
Detach `Vector`s from MATLAB

willGraham01:
# Description

* Continues #285 | Resolves `FieldComponentsVector` and `FrequencyExtractVector`.

Detaches the `FieldComponentsVector` and `FrequencyExtractVector` classes from the `Vector` class, which depended on MATLAB. 
- These two classes have been reduced to `typedef`s of `std::vector`.

The methods these classes had are now functions inside a (new) namespace `tdms_vector_utils` within `utils.h`.
- A lot of our existing classes have custom methods that perform identical tasks to those that these functions perform. Running a pass over our classes and refactoring those methods might give us a better idea of what should be a "class/struct" for us and what can just be `typedef`ed.

## Testing

Changes have been made in such a way as to be compatible with the existing tests, and not to alter the expected behaviour.

`test_utils.h` does have an additional test for the new `to_vector_int` method which we will need for casting data read from MATLAB files to `std::vector<int>`.


samcunliffe:
> Windows build does not support uint parallel for loop variables 😠

*sigh*. Might be able to isolate the Windows problem with a macro like `#if OpenMP_VERSION < 3`. Working over on [my fork](https://github.com/samcunliffe/TDMS/actions/runs/5179435835) to see if we can get it working.

tdms/include/hdf5_io/hdf5_reader.h
```diff
@@ -6,11 +6,13 @@
...
 #include "arrays/tdms_matrix.h"
+#include "arrays/vector_typedefs.h"
 #include "interface.h"
 
 /**
  * @brief Class wrapper of the reading of HDF5 format files.
- * @details Opens files in readonly and retrieves the datasets (in our case
+ * @details Opens files in readonly and retrieves the datasets (in our
+ * case
  *          **double, but can be anything in general).
```
samcunliffe, line 16:
```suggestion
 *          case, **double, but can be anything in general).
```

tdms/include/simulation_manager/objects_from_infile.h
```diff
@@ -72,6 +73,9 @@ class IndependentObjectsFromInfile {
 
   GridLabels input_grid_labels;//!< cartesian labels of the Yee cells
 
+  FrequencyExtractVector
+          f_ex_vec;//< Vector of frequencies to extract field & phasors at
+
```
samcunliffe, line 78:
```suggestion
  //! Vector of frequencies to extract field & phasors at
  FrequencyExtractVector f_ex_vec;

```

samcunliffe: (I hope that also makes clang-format happy.)

tdms/include/utils.h
```diff
@@ -2,7 +2,12 @@
  * @file utils.h
  * @brief Useful miscellaneous utility functions
  */
+#pragma once
```
samcunliffe, line 5: 😱

samcunliffe (approved):
I like this very much.

(Except for the coverage weirdness that we discussed IRL. I do not like that.)

samcunliffe:
🚀

PR #330, opened 2023-06-02
`HDF5Base` can identify empty objects

willGraham01:
# Context

Naturally, MATLAB does not save empty arrays or structs as objects with no elements or zero-size, instead it saves them as 2-by-1 arrays with 0's populating the data. This means that a simple comparison against the number of elements or members does not provide the correct information when attempting to determine whether an empty input has been passed.

MATLAB _does_ however attach an attribute to any object that it marks as empty: `MATLAB_empty`. This `uint8` is set to 1 if the object is indeed an empty array, so _that_ is what we will use to distinguish.

* Concerns #70 
* Overcomes some conversion problems faced by #285 (essentially replacing the `mxIsEmpty` function).

## Details

Adds the `HDF5Base::flagged_MATLAB_empty` method to replace the `mxIsEmpty` function, which a number of our classes depend on when reading from the input file.

Going forward, if the input generation is ever moved away from MATLAB too, a workaround will need to be decided to replace/handle `.hdf5` files that are not created with MATLAB, as the analogue of this flag will not exist.

## Testing

- `tdms/tests/unit/hdf5_io/test_hdf5_base.cpp` added with extensive tests for the new `flagged_MATLAB_empty` method.
- (The scripts that produce the) `.mat` and `.hdf5` unit test data has been updated with the necessary additions


tdms/include/hdf5_io/hdf5_base.h
```diff
@@ -90,4 +90,22 @@ class HDF5Base {
...
+   * populating the data. This means that a simple comparison against the
+   * number of elements or members does not provide the correct information when
+   * attempting to determine whether an empty input has been passed.
+   *
+   * MATLAB _does_ however attach an attribute to any object that it marks as
+   * empty, MATLAB_empty. This uint8 is set to 1 if the object is indeed an
+   * empty array, so _that_ is what we will use to distinguish.
+   * @param object_path Path to the object under file root; /object_path.
+   * @return true The object is flagged as being the empty MATLAB array.
+   * @return false Otherwise.
```
samcunliffe, line 108:
```suggestion
   * @brief Returns true if the object under /object_path is flagged as
   *        MATLAB_empty.
   *
   * @details Naturally, MATLAB does not save empty arrays or structs as objects
   *          with no elements or size, instead it saves them as 2-by-1 arrays
   *          with 0's populating the data. A simple comparison against the
   *          number of elements or members does not provide the correct
   *          information when attempting to determine if empty input has been
   *          passed.
   *
   *          MATLAB _does_, however. attach an attribute to any object that it
   *          marks as empty: MATLAB_empty. This is a uint8 set to 1 if the
   *          object is an empty array, so use that to distinguish.
   *
   * @param object_path Path to the object under file root; /object_path.
   * @return true The object is flagged as being the empty MATLAB array.
   * @return false Otherwise.
 ```
[ Minor simplification + alignment. ]

File: tdms/src/hdf5_io/hdf5_base.cpp [hunk 2]
```diff
@@ -49,8 +50,58 @@ H5Dimension HDF5Base::shape_of(const std::string &group_name,
...
 
 bool HDF5Base::is_ok() const {
-  // TODO: check for file health might be unnessicary given we've constructed
+  // TODO: check for file health might be unnecessary given we've constructed
   // the object.
   return file_->isHdf5(filename_);
   // return file_->isAccessible(filename_) && file_->isHdf5(filename_);
 }
+
+bool HDF5Base::flagged_MATLAB_empty(const std::string &object_path) const {
```
samcunliffe, line 59: Maybe `is_flagged_as_MATLAB_empty(object_path)`?
I don't insist though.

tdms/src/hdf5_io/hdf5_base.cpp
```diff
@@ -49,8 +50,58 @@ H5Dimension HDF5Base::shape_of(const std::string &group_name,
...
+  // Can't check anything if there's no file
+  if (file_ == nullptr) { throw std::runtime_error("No file opened"); }
```
samcunliffe, line 61: I'd use our internal healthcheck:
```suggestion
  if (!is_ok()) { throw std::runtime_error("Problem with the file!"); }
```
And put the `file_ != nullptr` check up in `HDF5Base::is_ok`.

Although by opening the `H5::File` in the constructor, we're guaranteeing that `file_` can never be null (and probably can never _not_ be `isHdf5`).

samcunliffe: If we delete explicitly the default constructor, I think we're safe.

That is: it never makes sense to have a `HDF5Base` without a `file_` which needs a `path` and `mode` which means our constructor _must_ provide these.

... what do you think? (Not for this PR)

tdms/src/hdf5_io/hdf5_base.cpp
```diff
@@ -49,8 +50,58 @@ H5Dimension HDF5Base::shape_of(const std::string &group_name,
...
+  if (!file_->exists(object_path)) {
+    throw std::runtime_error(filename_ + " has no object " + object_path);
+  }
+  hid_t object_reference = file_->getObjId(object_path);
+
+  // The object could be a group or a dataset, so we need to account for this
+  H5I_type_t object_type = file_->getHDFObjType(object_reference);
+  if (object_type == H5I_GROUP) {
+    // Dealing with a group
+    H5::Group object = file_->openGroup(object_path);
```
samcunliffe, line 76:
```suggestion
  // Attempt to fetch the object requested
  if (!file_->exists(object_path)) {
    throw std::runtime_error(filename_ + " has no object " + object_path);
  }
  hid_t object_reference = file_->getObjId(object_path);

  H5I_type_t object_type = file_->getHDFObjType(object_reference);
  H5::Attribute empty_attribute;  // will point to the MATLAB_empty attribute
  if (object_type == H5I_GROUP) {
      H5::Group object = file_->openGroup(object_path);
```
(Move declaration of `empty_attribute` down to just before it's used and delete some comments that describe some self-describing code.)

That is to say:
```C++
int thing_named_very_well;
if (thing_named_very_well == SOME_KIND) {
    // ...
} else if (thing_named_very_well == SOME_OTHER_KIND) {
    // ...
}
```
doesn't need a comment saying `// figure out what kind of thing and do ... accordingly`. I can already see that because your code is good.

tdms/tests/unit/benchmark_scripts/create_hdf5_test_file.py
```diff
@@ -16,12 +16,11 @@ def create_hdf5_test_file() -> None:
...
     file = h5py.File(FNAME_TO_CREATE, "w")
 
-    # Create a group under root
-    read_in_test = file.require_group("read_in_test")
-
     # Create test data to read in
     consecutive_numbers = np.arange(0, 12, dtype=float)
 
+    # Create a group under root
+    read_in_test = file.require_group("read_in_test")
```
samcunliffe, line 23: This, however, _does_ need a comment because it's weird to me that `require_group` _creates_ a group.

samcunliffe: #libhdf5-world-problems

samcunliffe (approved):
🍉 

No substantial comments. You can take or leave these...

tdms/src/hdf5_io/hdf5_base.cpp
willGraham01: Sounds sensible - the default constructor is already `protected` anyway so we can't actually use it to instantiate `HDF5Base` nor any of the derived classes.

tdms/tests/unit/benchmark_scripts/create_hdf5_test_file.py
willGraham01: That's `h5py` for you :sweat_smile:

PR #339, opened 2023-06-06
Add the `_OPENMP` version (date of spec) to `tdms --version`

samcunliffe:
As title. `tdms --version` is called in our CI so we should be able to get what [version of the OpenMP spec](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenMP.cmake#L435). VisualStudio is obeying the Windows runners.

Up to you if we actually want to spit the OpenMP version spec out with `tdms --version`. I'd say it doesn't hurt.

### DIFF
--- a/tdms/src/argument_parser.cpp
+++ b/tdms/src/argument_parser.cpp
@@ -57,6 +57,7 @@ void ArgumentParser::print_help_message() {
 
 void ArgumentParser::print_version() {
   fprintf(stdout, "TDMS version: %s\n", tdms::VERSION.c_str());
+  fprintf(stdout, "OpenMP version: %i\n", _OPENMP);
 }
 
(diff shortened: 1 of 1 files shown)


willGraham01 (approved):

samcunliffe:
And the results are in:

* Windows: 200203 (2.0)
* Ubuntu: 201511 (4.5)
* MacOS: 201811 (5.0)

Where the parentheses are my interpretation of the [mapping of spec date to version](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenMP.cmake#L435). So we probably just want a 

```
#if (_OPENMP < 200805)
    // unsigned to signed conversion because: VS
#else
   // normal operating systems
#endif
```

PR #343, opened 2023-06-06
Read complex values from `.mat` files via HDF5 C++ API

willGraham01:
Allows us to read complex MATLAB arrays via the HDF5 API.

To do this, we have to define (at runtime) a `H5::CompType` instance that describes the data-structure that needs to be read in - in this case, that describes the format in which MATLAB saves complex data. Note that there is no HDF5 standard for complex datatypes, so trying to read complex data from something like `h5py`- created files will require separate, but similar, treatment.

- Concerns #285 | We can now read in complex data from .mat input files

## Implementation

The `hdf5_compound_types.{h,cpp}` file has been created to contain compound types.

Currently there is only one compound type, `MATLAB_complex`. But the following convention has been set given the remark above about compound datatypes and the lack of standard format for complex-values in particular:
- Each compound type has it's own namespace within this file
- Each namespace contains:
  - The `struct` that is compatible with the compound data-type to be read
  - A `to_hdf5_CompType()` method that returns a `H5::CompType` instance, that can be passed to the API to read data into the aforementioned `struct` via the `read` method.
  - Any number of helpful converter functions to `std::` datatypes (or `std::vector`s thereof) so that we don't have to reimplement algebra.

## Testing

Additional tests in `test_hdf5_compound_types.cpp` that check that:
- The new overload of the `HDF5Reader::read()` method performs as expected for `MATLAB_complex` types
- The other supporting functions in the namespace perform correctly.


samcunliffe (approved):

samcunliffe:
Nice!

PR #345, opened 2023-06-16
Also run release builds on CI

samcunliffe:


samcunliffe:
@willGraham01 not actually sure if we want all of this, but if development is slowing down, then we can at least keep track that all of these builds work.

willGraham01 (approved):

PR #351, opened 2024-11-18
Update `upload-artifact`, `download-artifact` and `checkout` versions.

samcunliffe:
Some maintenance. [Version 3 is deprecated](https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/#artifacts-v3-brownouts).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
 
     steps:
       - name: Check out repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Set up MATLAB
         uses: matlab-actions/setup-matlab@v1.2.3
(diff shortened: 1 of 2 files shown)


samcunliffe:
MATLAB vs. ARM MacOS.

willGraham01:
MATLAB still not happy - this an upstream issue I presume?

willGraham01:
> So yeah. We could try going back a few versions of the action...?

Don't think that will work, your [first commit](https://github.com/UCL/TDMS/pull/351/commits/e60ce9ddbaa0d96dd4b1384d63dc53cc66a95e22) didn't bump the `setup-matlab` version and ran into the same problem as the last one.

Think we need to wait for the fix, or would have to drop the MacOS tests otherwise.

.github/workflows/ci.yml
```diff
@@ -8,28 +8,27 @@ name: Build and test
...
 
-
 jobs:
   linting:
     name: Linting checks with pre-commit
     runs-on: ubuntu-latest
     steps:
-      - uses: ucl/composite-actions/pre-commit@v1
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
```
samcunliffe, line 30:
```suggestion
      - uses: actions/setup-python@v4
      - name: Check clang-format version
         run: clang-format --version
```

.github/workflows/ci.yml
```diff
@@ -8,28 +8,31 @@ name: Build and test
...
 # changed).
 
-
 on:
   push:
     branches: ["main"]
   pull_request:
   merge_group:
+  schedule:
+    - cron: "0 9 15 * *"
```
samcunliffe, line 17: Since the build is brittle, I propose we run it monthly to keep vaguelly on top of things...?

willGraham01 (approved):

